PR #3, opened 2021-11-01
Create empty objects along json path if needed

zhan-airplane:
This does a slightly more general version of what I originally set out to do - basically, if any string component of a path we're trying to walk down evaluates to `null`, we insert an empty object at that point and carry on walking down (unless it's the very end of an append, in which case an empty array is created)

---
cc @colinking


pkg/outputs/apply.go
```diff
@@ -112,70 +75,20 @@ func applyAppend(jsPath string, v interface{}, o *ojson.Value) error {
...
-				cur = arr[c]
-			}
-
-		default:
-			return errors.New("unexpected component type")
-		}
+	var locArr []interface{}
+	locVal := getAtLocation(loc)
+	if locVal == nil {
+		locVal = updateLocation(loc, []interface{}{})
```
joshma, line 86: could use a comment - is this to handle an array instead of object as the default?

although, why does this have to exist? can `updateLocation` handle this around L78?

pkg/outputs/location.go
```diff
@@ -0,0 +1,97 @@
...
+		loc.Root.V = v
+	} else if loc.Obj != nil {
+		loc.Obj.Obj.Set(loc.Obj.Key, v)
+	} else if loc.Arr != nil {
+		(*loc.Arr.Arr)[loc.Arr.Key] = v
+	}
+	return v
+}
+
+func getLocation(p path.P, root *ojson.Value) (location, error) {
```
joshma, line 50: I think this could use a docstring

```
// getLocation returns a location at path p, usually for modifying.
// This will set empty objects / arrays along the path to the location so that...
```

pkg/outputs/apply.go
zhan-airplane: Actually yeah you're right, the other `updateLocation` is good enough. I'll add a comment too though

joshma (approved):

PR #4, opened 2021-11-13
Pull out build code from CLI

leeweisberger:
Pull build code to a common place so that both the CLI and API can reference it.

The entire pr is copy/paste

---

This isn't super important - mostly just did this as a way to poke through how builds work in the CLI.


joshma:
@colinking can I get your eyes on this instead? I think it looks fine but would feel better if you did a pass.

pkg/build/types.go
```diff
@@ -0,0 +1,32 @@
+package build
+
+type KindOptions map[string]interface{}
```
colinking, line 3: Sweet, so this is pulled out of `api`? https://sourcegraph.com/github.com/airplanedev/airport@f4a4317f51b64b8f1300f2ca279a379738cb22ab/-/blob/pkg/builder/builder.go?L294

Nice

colinking (approved):
LGTM! Let's remove `examples` from `cli` and have `runtime` use these instead, so they don't get duplicated.

PR #7, opened 2021-11-18
Copy changes that were made in cli in build code

leeweisberger:
Before removing cli/build, copy changes that were made in the last few days over to the lib.


.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,11 @@
+# Contributing to the Airplane lib
```
leeweisberger, line 1: Added this guy. Unrelated

pkg/build/builder.go
```diff
@@ -282,7 +282,7 @@ func (b *Builder) authconfigs() map[string]types.AuthConfig {
 //
 // The following string manipulations won't matter for non-ksuid
 // IDs (the current scheme).
-func sanitizeTaskID(s string) string {
+func SanitizeTaskID(s string) string {
```
leeweisberger, line 285: Also an unrelated change - exporting this b/c the CLI uses it

pkg/build/types.go
```diff
@@ -17,9 +17,6 @@ const (
 	TaskKindREST TaskKind = "rest"
 )
 
-// Type enumerates parameter types.
```
leeweisberger, line 20: nobody is using this

pkg/build/node.go
```diff
@@ -35,20 +36,40 @@ func node(root string, options KindOptions) (string, error) {
 
 	workdir, _ := options["workdir"].(string)
 	pathPackageJSON := filepath.Join(root, "package.json")
+	hasPackageJSON := fsx.AssertExistsAll(pathPackageJSON) == nil
 	pathYarnLock := filepath.Join(root, "yarn.lock")
+	var usesYarnWorkspaces bool
```
colinking, line 41:
```suggestion
	var usesWorkspaces bool
```

pkg/build/node.go
```diff
@@ -104,38 +125,37 @@ func node(root string, options KindOptions) (string, error) {
...
 		RUN npm install -g typescript@4.2 && \
 			npm install -g esbuild@0.12 --unsafe-perm
-		COPY . /airplane
-
 		RUN mkdir -p /airplane/.airplane && \
 			cd /airplane/.airplane && \
 			{{.InlineShimPackageJSON}} > package.json && \
 			npm install
-
+		COPY package*.json yarn.* /airplane/
```
colinking, line 144: I get `package*.json` (to match `package.json` and `package-lock.json`). But why `yarn.*`? Isn't there just `yarn.lock`? Or is there a yarn workspaces file?

pkg/build/node.go
```diff
@@ -35,20 +36,40 @@ func node(root string, options KindOptions) (string, error) {
 
 	workdir, _ := options["workdir"].(string)
 	pathPackageJSON := filepath.Join(root, "package.json")
+	hasPackageJSON := fsx.AssertExistsAll(pathPackageJSON) == nil
 	pathYarnLock := filepath.Join(root, "yarn.lock")
+	var usesYarnWorkspaces bool
+
+	if hasPackageJSON {
```
colinking, line 43: Just realized, could you add a comment here about why we do this special casing? 🙏🏻

colinking (approved):

pkg/build/node.go
leeweisberger: ah the yarn.* is a trick that will copy yarn.* if it exists and ignore if it doesn't. If you just do yarn.lock, it errors if there is no yarn.lock file to copy - we could conditionally copy the yarn file behind another if/else, but thought this was nicer

PR #9, opened 2021-11-23
Loader fixes

leeweisberger:
1. Support not showing the loader (will use this to hide the loader in debug mode)
2. Loader handles concurrency
3. Loader does not show in non terminals like the log output of github actions (didn't test this, but I think it will work)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 	github.com/moby/sys/mount v0.3.0 // indirect
 	github.com/sabhiram/go-gitignore v0.0.0-20210923224102-525f6e181f06
 	github.com/segmentio/ksuid v1.0.4
-	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2
+	golang.org/x/sys v0.0.0-20211117180635-dee7805ff2e1 // indirect
+	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -10,34 +10,10 @@ require (
 )
 
 require (
-	github.com/Azure/go-ansiterm v0.0.0-20170929234023-d6e3b3328b78 // indirect
```
colinking, line 13: Are these changes just from `go mod tidy`?

pkg/build/logger/logger.go
```diff
@@ -64,29 +66,61 @@ func Debug(msg string, args ...interface{}) {
...
-	return &Loader{spin: spinner.New(spinner.CharSets[11], 100*time.Millisecond, spinner.WithWriter(os.Stderr))}
+// NoopLoader doesn't do anything.
+type NoopLoader struct {
+}
+
+type LoaderOpts struct {
+	HideLoader bool
+}
+
+func NewLoader(opts LoaderOpts) Loader {
```
colinking, line 89: Ah, feels off that the loader is in our `lib/build` package since it's so specific to the`cli`. Hopefully over time we can better separate the two!

colinking (approved):

pkg/build/logger/logger.go
leeweisberger: Agreed. The logger shouldn't even be here. closing in favor of https://github.com/airplanedev/cli/pull/330

leeweisberger:
https://github.com/airplanedev/cli/pull/330

PR #10, opened 2021-12-06
Move runtime to lib

leeweisberger:
Ultimate goal here is to be able to reuse a lot of this code in the GitHub app. We really want the CLI to just do CLI things and move all of the business logic to lib/

1. Move all runtime code to lib so we can reuse it.
2. Move examples to lib. No longer needed in CLI.
3. Remove logger from lib

To accomplish 1 - I copied over cli/runtime. The only changes I made are
* Added a `logger` param to `PrepareRun` so we can log things from within
* Rather than taking a full api.Task, `Generate` and `Comment` just take the values that they need


pkg/build/builder.go
```diff
@@ -155,10 +154,8 @@ func (b *Builder) Build(ctx context.Context, taskID, version string) (*Response,
 	if err != nil {
 		return nil, errors.Wrap(err, "creating dockerfile")
 	}
-	logger.Debug(strings.TrimSpace(dockerfile))
```
leeweisberger, line 158: I removed a few debug logs from within build code. I want to get rid of the debugger in lib, and these logs don't seem worth plumbing a logger through

pkg/runtime/comment.go
```diff
@@ -0,0 +1,84 @@
...
+	commentRegex = regexp.MustCompile(`Linked to (https://.*air.*/t/.*) \[do not edit this line\]`)
+	// maxBytesToReadForSlug is the max bytes we should read in a file when looking for a task slug.
+	maxBytesToReadForSlug int64 = 4096
+)
+
+// Comment generates a linking comment that is used
+// to associate a script file with an Airplane task.
+//
+// This comment can be parsed out of a script file using Slug.
+func Comment(r Interface, taskURL string) string {
```
leeweisberger, line 26: One change - this used to take an api.Task. Now it just takes a task url

pkg/runtime/javascript/javascript.go
```diff
@@ -0,0 +1,276 @@
...
+// Data represents the data template.
+type data struct {
+	Comment string
+}
+
+// Runtime implementaton.
+type Runtime struct{}
+
+// Generate implementation.
+func (r Runtime) Generate(t *runtime.Task) ([]byte, fs.FileMode, error) {
```
leeweisberger, line 58: Second change - this used to take an api.Task. Now it takes a much simpler runtime.Task

pkg/runtime/types.go
```diff
@@ -0,0 +1,39 @@
+package runtime
+
+// Task represents a task.
+type Task struct {
```
leeweisberger, line 4: Here is the trimmed down version of task we are using

pkg/utils/handlebars/handlebars.go
```diff
@@ -0,0 +1,20 @@
+// Light wrapper around handlebars.
```
leeweisberger, line 1: copied this util from cli as well

pkg/utils/prompt.go
```diff
@@ -0,0 +1,30 @@
+package utils
```
leeweisberger, line 1: copied this util from cli

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// KindOptions specifies any runtime-specific task configuration.
+	KindOptions build.KindOptions
+}
+
+type Logger interface {
+	Log(msg string, args ...interface{})
+	Warning(msg string, args ...interface{})
+	Debug(msg string, args ...interface{})
+}
```
leeweisberger, line 105: I added this interface so the CLI can pass its logger in

colinking:
@leeweisberger can you add `node_modules` to `.gitignore`?

colinking (approved):

PR #12, opened 2021-12-10
Gracefully handle missing airplanesdk in shim

joshma:
Rather than make airplanesdk a hard dependency, this would instead print
the `ret` value to stdout and print a warning to stderr.

This way, new users can `airplane dev` their scripts without any
packages being installed.


joshma:
and cc @colinking

leeweisberger (approved):

PR #16, opened 2022-01-06
Move reusable deploy code from cli -> lib

leeweisberger:
Move three bits of code so that we can reuse them elsewhere for deploys outside of cli:
- Task definition
- Task discoverer
- Task archiver

This is a copy paste except where otherwise noted in inline comments


pkg/deploy/api/client.go
```diff
@@ -0,0 +1,258 @@
...
+import (
+	"context"
+	"encoding/json"
+	"time"
+
+	"github.com/airplanedev/lib/pkg/build"
+	"gopkg.in/yaml.v3"
+)
+
+type APIClient interface {
```
leeweisberger, line 12: The deploy code uses some api calls. This is the most annoying part of this migration.

Rather than create an implementation of an api client, any consumers of the deploy code have to pass an interface that satisfies these few api functions.

Preferably we wouldn't have this dependency, but it will be a bit disruptive to factor it out. Rather than thinking of this as an actual api client, I think it's better to think of it like an interface that might be satisfied by an api client. We just need a way to get a task given a slug e.g.

pkg/deploy/discover/defn_discover.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/airplanedev/lib/pkg/deploy/taskdir/definitions"
+	"github.com/airplanedev/lib/pkg/utils/logger"
+)
+
+type DefnDiscoverer struct {
+	Client             api.APIClient
+	AssumeYes          bool
+	AssumeNo           bool
+	Logger             logger.Logger
+	MissingTaskHandler func(context.Context, definitions.DefinitionInterface) (*api.Task, error)
```
leeweisberger, line 18: I added this extra piece of config - before this was hardcoded in the defndiscoverer to prompt the user to create a new task if one was missing. However this behavior is pretty cli specific, so instead this discoverer takes an input that the cli can call to prompt the user. Other consumers might not pass this in

pkg/runtime/javascript/javascript.go
```diff
@@ -18,6 +18,7 @@ import (
 
 	"github.com/airplanedev/lib/pkg/runtime"
 	"github.com/airplanedev/lib/pkg/utils/fsx"
+	"github.com/airplanedev/lib/pkg/utils/logger"
```
leeweisberger, line 21: I moved the logger interface to a more common place. That's the only change outside of the copy/pasted `discover/` folder

francescaleung (approved):

pkg/deploy/api/client.go
```diff
@@ -0,0 +1,278 @@
+package api
```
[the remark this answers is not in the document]
leeweisberger: I was thinking the same thing... moved

PR #19, opened 2022-01-08
Add mock archiver. Init runtimes

leeweisberger:
There are two changes in this PR:
1. Add a mock archiver so users don't have to define their own mock.
2. Initialize runtimes in the ScriptDiscoverer - without this change, users of the discoverer are expected to initialize the runtimes. I generally don't like this init pattern, but it's definitely strange to expect users to import packages before using the discoverer. I also allow us to register runtimes multiple times - don't see any reason why we wouldn't allow this. If we keep the panic, this would be a breaking change since the CLI also inits the runtimes.


pkg/deploy/discover/discover_test.go
```diff
@@ -9,10 +9,6 @@ import (
 	"github.com/airplanedev/lib/pkg/api/mock"
 	"github.com/airplanedev/lib/pkg/build"
 	"github.com/airplanedev/lib/pkg/deploy/taskdir/definitions"
-	_ "github.com/airplanedev/lib/pkg/runtime/javascript"
```
leeweisberger, line 12: We no longer need to do this in the test.

pkg/runtime/runtime.go
```diff
@@ -106,9 +105,6 @@ var runtimes = make(map[string]Interface)
 
 // Register registers the given ext with r.
 func Register(ext string, r Interface) {
-	if _, ok := runtimes[ext]; ok {
-		panic(fmt.Sprintf("runtime: %s already registered", ext))
-	}
```
francescaleung, line 111: hmm, this still seems useful to have around. Any way we could keep it?

leeweisberger: I added back. The init only gets called once (I thought it met get called once per import). Did some reading and manually tested.

francescaleung (approved):

PR #21, opened 2022-01-11
Add script for populating public cache

colinking:
Added a script to make it easier to publish to our public cache. Needed this for our staging environment.

The deno images now have official mirrors, so I moved to that. The old ones no longer exist.

I didn't set up gcloud access on the Airplane task, so for now this task is only run locally.


leeweisberger (approved):

pkg/deploy/discover/fixtures/nestedScripts/package.json
```diff
@@ -0,0 +1 @@
+{}
```
colinking, line 1: cc @leeweisberger because of the new package.json in the root directory, I had to add `package.json` files here to fix the task discovery test above

PR #22, opened 2022-01-12
Breaking changes - Speed up node installations

leeweisberger:
These are breaking changes - if users have everything set up correctly, nothing will break. Things will only break if:
- Users are relying on dev dependencies for their production build

Benefits of these changes:
- Faster installs and smaller containers - no need to install useless dev dependencies in production.
- Don't auto upgrade packages - we should not be auto-upgrading packages during our prod build. Today, we might do an upgrade past what the user has installed via their package.lock or yarn.lock

It looks like we explicitly have a test for dev dependencies - what's the reasoning behind this? Is it just to avoid breaking changes? Maybe we can add an option to include dev dependencies if the user desires?


pkg/build/node.go
```diff
@@ -155,9 +158,11 @@ func node(root string, options KindOptions) (string, error) {
 		{{end}}
 
 		{{if .IsYarn}}
-		RUN yarn --non-interactive
+		RUN yarn --non-interactive --production --frozen-lockfile
+		{{else if .HasPackageLock}}
+		RUN npm ci --production
 		{{else}}
-		RUN npm install
+		RUN npm i --production
```
joshma, line 165: can we use the full `npm install` word?

pkg/build/node.go
```diff
@@ -155,9 +158,11 @@ func node(root string, options KindOptions) (string, error) {
 		{{end}}
 
 		{{if .IsYarn}}
-		RUN yarn --non-interactive
+		RUN yarn --non-interactive --production --frozen-lockfile
+		{{else if .HasPackageLock}}
+		RUN npm ci --production
```
joshma, line 163: Maybe add a comment like

```
// Use npm ci if possible, since it's faster and behaves better:
// https://docs.npmjs.com/cli/v8/commands/npm-ci
```

pkg/build/node.go
```diff
@@ -155,9 +158,11 @@ func node(root string, options KindOptions) (string, error) {
 		{{end}}
 
 		{{if .IsYarn}}
-		RUN yarn --non-interactive
+		RUN yarn --non-interactive --production --frozen-lockfile
```
joshma, line 161: minor but while you're here, this seems clearer:

```suggestion
		RUN yarn install --non-interactive --production --frozen-lockfile
```

joshma (approved):
I _think_ this is good but want to check @colinking 's take as well. I vaguely remember dev deps being useful if e.g. you have a version of typescript you want to use? And build deps would be in dev? But we're using `esbuild` here and wouldn't end up calling that.

leeweisberger:
I actually might close this and make these optional. Maybe not best to introduce something by default that breaks flows unexpectedly

colinking (approved):
We could do a multi-stage build with a build stage that installs dev deps (and e.g. runs future user-configurable build commands) and a main stage that only includes prod deps. This would allow us to remove other build deps like typescript/esbuild from the main stage, plus uncompiled code.

+1 on locking exact deps

joshma:
@leeweisberger to stress-test our monitoring, can you track node builds and confirm that they go down in time? I guess it's only after we merge this into `airport` right?

PR #24, opened 2022-01-12
Add NewDefinitionFromTask_0_3 function

francescaleung:
With tests! Also added tests for Definition_0_3 -> UpdateTaskRequest function.

Resolves AIR-2577.


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -168,6 +198,16 @@ func (d *GoDefinition_0_3) fillInUpdateTaskRequest(ctx context.Context, client a
 	return nil
 }
 
+func (d *GoDefinition_0_3) hydrateFromTask(ctx context.Context, client api.IAPIClient, t *api.Task) error {
+	d.Arguments = t.Arguments
+	if v, ok := t.KindOptions["entrypoint"]; ok {
+		if sv, ok := v.(string); ok {
```
leeweisberger, line 204: what if it's not a string? is that an error case? or expected?

leeweisberger: Same in the other places

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -789,8 +930,119 @@ func (d *Definition_0_3) GetSlug() string {
...
+		}
+
+		switch param.Type {
+		case "string":
+			switch param.Component {
+			case api.ComponentTextarea:
+				p.Type = "longtext"
+			case api.ComponentEditorSQL:
+				p.Type = "sql"
+			default:
```
leeweisberger, line 977: we might want to error on default in case we add another option? Unless there is no other way to differentiate shorttext (or if shorttext is the default, i'm unsure)

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -789,8 +930,119 @@ func (d *Definition_0_3) GetSlug() string {
...
+			p.Required = &v
+		}
+
+		p.Regex = param.Constraints.Regex
+
+		if len(param.Constraints.Options) > 0 {
+			p.Options = make([]OptionDefinition_0_3, len(param.Constraints.Options))
+			for j, opt := range param.Constraints.Options {
+				p.Options[j].Label = opt.Label
+				p.Options[j].Value = opt.Value
```
leeweisberger, line 997: nit: could just do this in one line p.Options[j] = OptionDefinition_0_3{

leeweisberger (approved):
nice tests!

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: you're right, we should explicitly check for the shorttext case & error on default.

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: hmm it's probably an error case

leeweisberger: Let's throw an error then because if we don't, we will silently fail - the user will think their task def is valid and working but we're actually silently skipping fields!

leeweisberger: Although we should already hit this error when we convert the yaml -> go.... but I think we should add errors here anyways

PR #27, opened 2022-01-13
Add custom install and postinstall hooks

leeweisberger:
Add custom hooks accessible as:
```
airplane: {
  "install":
  "postinstall":
}
```

Install allows you to specify a custom installation script. e.g. if you want to install dev dependencies or do something else custom.

PostInstall allows you to run commands after install and before the build. e.g. generating a prisma client after the prisma cli has been installed.

Eventually I think we might want to (additionally) surface this in the deploy configuration that @rose-huang is working on.


pkg/build/node.go
```diff
@@ -130,6 +141,7 @@ func node(root string, options KindOptions) (string, error) {
 	// in which case we could introduce an extra step for performing build commands.
 	return applyTemplate(heredoc.Doc(`
 		FROM {{.Base}}
+		ENV NODE_ENV=production
```
leeweisberger, line 144: Might as well add this. Used by different node frameworks to optimize production builds

pkg/build/node.go
```diff
@@ -157,22 +169,21 @@ func node(root string, options KindOptions) (string, error) {
...
+
+		{{if .PostInstallCommand}}
+		RUN {{.PostInstallCommand}}
+		{{end}}
 		
 		RUN {{.InlineShim}} > /airplane/.airplane/shim.js && \
 			esbuild /airplane/.airplane/shim.js \
 				--bundle \
 				--external:airplane \
+				--minify \
```
leeweisberger, line 186: Another unrelated change. Slightly smaller builds.

pkg/runtime/runtime.go
```diff
@@ -32,11 +32,6 @@ var (
 	ErrNotImplemented = errors.New("runtime: not implemented")
 )
 
-// Settings represent Airplane specific settings.
-type Settings struct {
```
leeweisberger, line 36: circular dep issues. Moved to /build

pkg/build/node.go
```diff
@@ -157,22 +169,21 @@ func node(root string, options KindOptions) (string, error) {
...
 		{{end}}
 
-		{{if .IsYarn}}
-		RUN yarn --non-interactive --production --frozen-lockfile
-		{{else if .HasPackageLock}}
-		RUN npm ci --production
-		{{else}}
-		RUN npm i --production
-		{{end}}
+		RUN {{.InstallCommand}}
```
colinking, line 172: What's the use-case for the custom install command?

pkg/build/node.go
```diff
@@ -115,6 +114,18 @@ func node(root string, options KindOptions) (string, error) {
 	}
 	cfg.InlineShim = inlineString(shim)
 
+	installCommand := "npm install --production"
+	if pkg.Settings.InstallCommand != "" {
+		installCommand = pkg.Settings.InstallCommand
```
colinking, line 119: Could you `s/\n/\\n`? Otherwise folks will hit issues if they try to add new lines as they'll get interpreted as Dockerfile commands.

pkg/build/node.go
colinking: This will effect error stack traces, since they'll reference minified function names

pkg/build/node.go
colinking: Is the idea that Prisma users would override this with:

```
cp prisma . && yarn install --non-interactive --frozen-lockfile
```

Basically (1) copy over whatever files are needed for `prisma generate` then (2) override our install command to install dev deps since `@prisma/client` will be a dev dep?

My understanding is they would not need a `postinstall` command since the prisma client runs `prisma generate` from a standard `package.json` `postinstall` hook.

Given all this, could you add a test case that uses Prisma? I agree it'll be a common Airplane use-case.

pkg/examples/typescript/yarn/package.json
```diff
@@ -2,4 +2,5 @@
   "dependencies": {
     "airplane": "^0.0.5"
   }
+
```
colinking, line 5: typo?

colinking (approved):

pkg/build/node.go
leeweisberger: ah good call. Let me revert this then. Not worth

pkg/build/node.go
leeweisberger: That was what I was thinking - they can override this to install dev deps.

They would also need a postinstall hook b/c when we run install, we don't have the rest of their code in the container yet (for caching purposes) so they'd need to manually run prisma generate.

I was also thinking we might want to auto-detect prisma and just do this correctly by default, but I think I'd prefer to add the ability to customize before we start auto-detecting

leeweisberger: Added a test here that shows the workflow: https://github.com/airplanedev/lib/pull/27/commits/adc5f6a40a1f83548935da4b019f6164c82d8a74

I will add this stuff to our documentation.

PR #28, opened 2022-01-13
Add permission conversion

francescaleung:
Resolves AIR-2547. Depends on https://github.com/airplanedev/airport/pull/1355.


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -859,10 +859,66 @@ func (d Definition_0_3) addParametersToUpdateTaskRequest(ctx context.Context, re
...
+	for _, member := range entities.Members {
+		membersByEmail[member.Email] = member
+	}
+
+	groupsByName := map[string]api.Group{}
+	for _, group := range entities.Groups {
+		groupsByName[group.Name] = group
+	}
+
+	for _, roleAndIdents := range []struct {
```
leeweisberger, line 883: can you store the roles and idents in a separate variable? Kinda hard to read the anonymous type + values

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -859,10 +859,66 @@ func (d Definition_0_3) addParametersToUpdateTaskRequest(ctx context.Context, re
...
+		roleID api.RoleID
+		idents []string
+	}{
```
leeweisberger, line 886: Just to think about - how could we auto-update this if we add more roles?

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1075,7 +1131,61 @@ func (d *Definition_0_3) convertTaskKindFromTask(ctx context.Context, client api
...
+	for _, group := range entities.Groups {
+		groupsByID[group.ID] = group
+	}
+
+	d.Permissions = &PermissionDefinition_0_3{}
+	for _, perm := range t.Permissions {
+		var ident string
+		if perm.SubUserID != nil {
+			if member, ok := membersByID[*perm.SubUserID]; !ok {
+				continue
```
leeweisberger, line 1158: What does this case represent? there is a permission but we don't know the user? Think this is an error case?

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1075,7 +1131,61 @@ func (d *Definition_0_3) convertTaskKindFromTask(ctx context.Context, client api
...
+			} else {
+				ident = member.Email
+			}
+		} else if perm.SubGroupID != nil {
+			if group, ok := groupsByID[*perm.SubGroupID]; !ok {
+				continue
```
leeweisberger, line 1164: same here

pkg/deploy/taskdir/definitions/def_0_3.go
leeweisberger: I guess for tasks at least this won't really grow

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -859,10 +859,66 @@ func (d Definition_0_3) addParametersToUpdateTaskRequest(ctx context.Context, re
...
+			idents: d.Permissions.Executers,
+		},
+		{
+			roleID: api.RoleTaskAdmin,
+			idents: d.Permissions.Admins,
+		},
+	} {
+		roleID := roleAndIdents.roleID
+		for _, ident := range roleAndIdents.idents {
+			if member, ok := membersByEmail[ident]; ok {
```
leeweisberger, line 906: So this would fail if there was a group with the same email as a member. Sounds fine. We should probably disallow this on member/group creation, but not a big deal

leeweisberger (approved):

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: moved to error case

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: yea, the case where a group name is the same as a user email isn't handled well. I'll add a ticket.

PR #30, opened 2022-01-14
Fix error handling in getDef, RunConstraints conversion

francescaleung:
The `IsEmpty` check ensures that we don't produce a yaml file that has the following:

```yaml
constraints:
  labels: null
```

which doesn't satisfy the schema.


pkg/deploy/discover/defn_discover.go
```diff
@@ -23,7 +23,7 @@ var _ TaskDiscoverer = &DefnDiscoverer{}
 func (dd *DefnDiscoverer) IsAirplaneTask(ctx context.Context, file string) (slug string, err error) {
 	def, err := getDef(file)
 	if err != nil {
-		return "", err
+		return "", nil
```
leeweisberger, line 26: Is this correct? This will skip the def silently if getDef errors out - this doesn't seem like what we want?

francescaleung: yea, I actually think this needs to be an IsTaskDef check instead

leeweisberger (approved):

PR #32, opened 2022-01-18
Use plugin to mark node modules as external

leeweisberger:
We have been marking all node modules as external to prevent them from being bundled by reading the package.json. However, when using workspaces, there might be multiple package.jsons. Any packages referenced here are being bundled - this causes us to ship two versions of these deps - one in node_modules and one that is bundled into the source file.

Instead, use a plugin to mark all node_module deps as external. Maintaining behavior where we bundle in esm modules and ancestors of esm modules.

---

Question - why aren't we bundling node_modules? Wouldn't it just be easier (and produce a lighter container) if we just bundled node modules? If we did that, we could remove some of the complexity here and we could just ship a container with the shim.js file which contains everything needed to run the application.


pkg/build/node.go
```diff
@@ -148,12 +135,6 @@ func node(root string, options KindOptions) (string, error) {
 		ARG BUILD_NPM_TOKEN
 		RUN [ -z "${BUILD_NPM_RC}" ] || echo "${BUILD_NPM_RC}" > .npmrc
 		RUN [ -z "${BUILD_NPM_TOKEN}" ] || echo "//registry.npmjs.org/:_authToken=${BUILD_NPM_TOKEN}" > .npmrc
-		# qemu (on m1 at least) segfaults while looking up a UID/GID for running
-		# postinstall scripts. We run as root with --unsafe-perm instead, skipping
-		# that lookup. Possibly could fix by building for linux/arm on m1 instead
-		# of always building for linux/amd64.
-		RUN npm install -g typescript@4.2 && \
```
leeweisberger, line 155: I don't think we need typescript? esbuild doesn't require it to build a typescript project.

pkg/build/node.go
```diff
@@ -148,12 +135,6 @@ func node(root string, options KindOptions) (string, error) {
 		ARG BUILD_NPM_TOKEN
 		RUN [ -z "${BUILD_NPM_RC}" ] || echo "${BUILD_NPM_RC}" > .npmrc
 		RUN [ -z "${BUILD_NPM_TOKEN}" ] || echo "//registry.npmjs.org/:_authToken=${BUILD_NPM_TOKEN}" > .npmrc
-		# qemu (on m1 at least) segfaults while looking up a UID/GID for running
-		# postinstall scripts. We run as root with --unsafe-perm instead, skipping
-		# that lookup. Possibly could fix by building for linux/arm on m1 instead
-		# of always building for linux/amd64.
```
leeweisberger, line 154: How do we test for this? I am no longer using --unsafe-perm. Want to know if I am breaking something. Do we just have to test this on an m1 macbook?

pkg/build/node-build.mjs
```diff
@@ -0,0 +1,49 @@
...
+          const moduleRegexp = /node_modules\/(.+?)\//
+          const moduleMatch = resolved.match(moduleRegexp);
+          const module = moduleMatch[1];
+          if (esmModules.has(module)) {
+            esmPaths.add(resolved)
+            return resolved
+          }
+
+          return {
+            external: true,
```
colinking, line 36: Do we need to pass other `resolved` fields back? e.g. `{...resolved, external: true}`

pkg/build/node-build.mjs
```diff
@@ -0,0 +1,49 @@
...
+const args = process.argv.slice(2)
+const nodeVersion = args[0];
+build({
+  plugins: [
+    // This plugin marks all dependencies as external so they are not bundled 
+    // unless they are esm modules or packages pulled in by esm modules.
+    NodeResolvePlugin({
+      extensions: [".ts", ".js"],
+      onResolved: (resolved, importer) => {
+        if (esmPaths.has(importer)) {
```
colinking, line 19: Is `importer` exactly `node-fetch`, or should we be doing a regex match like below?

pkg/build/node-build.mjs
```diff
@@ -0,0 +1,49 @@
...
+        return resolved;
+      },
+    }),
+  ],
+  entryPoints: ['/airplane/.airplane/shim.js'],
+  bundle: true,
+  platform: "node",
+  target: `node${nodeVersion}`,
+  outfile: "/airplane/.airplane/dist/shim.js",
+  external: ["airplane"],
```
colinking, line 48: I don't recall why airplane was external. Does it break anything if we bundle it?

pkg/build/node.go
```diff
@@ -180,12 +161,8 @@ func node(root string, options KindOptions) (string, error) {
 		{{end}}
 		
 		RUN {{.InlineShim}} > /airplane/.airplane/shim.js && \
-			esbuild /airplane/.airplane/shim.js \
-				--bundle \
-				--external:airplane \
-				--platform=node {{.ExternalFlags}} \
-				--target=node{{.NodeVersion}} \
-				--outfile=/airplane/.airplane/dist/shim.js
+			{{.InlineBuild}} > /airplane/.airplane/build.mjs && \
```
colinking, line 164: Make sure to test this with `airplane dev`, too: https://github.com/airplanedev/lib/blob/612dba0540cfbcc76a51a6ec41de3c1cda3d77d7/pkg/runtime/javascript/javascript.go#L209

pkg/build/node.go
colinking: I can give this a test for you.

joshma:
(removing my review from here to unblock)

PR #35, opened 2022-01-24
Add SQL runtime

francescaleung:
Adds a skeleton definition of SQL runtime without local execution. Also adds SQL + REST consideration to NeedsBuilding.

Resolves AIR-2485.

### DIFF
--- a/pkg/runtime/sql/sql.go
+++ b/pkg/runtime/sql/sql.go
@@ -1,0 +1,65 @@
+package sql
+
+import (
+	"context"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/airplanedev/lib/pkg/build"
+	"github.com/airplanedev/lib/pkg/runtime"
+	"github.com/airplanedev/lib/pkg/utils/logger"
+)
+
+// Init register the runtime.
+func init() {
+	runtime.Register(".sql", Runtime{})
+}
+
+// Code.
+var code = []byte(`-- Add your SQL code here:
+SELECT 1;
+
+-- Define SQL parameters in your definition file:
+-- SELECT * from users where user_id = :user_id;
+`)
+
+// Runtime implementation.
+type Runtime struct{}
+
+// PrepareRun implementation.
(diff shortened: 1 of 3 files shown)


pkg/runtime/sql/sql.go
```diff
@@ -0,0 +1,65 @@
...
+func init() {
+	runtime.Register(".sql", Runtime{})
+}
+
+// Code.
+var code = []byte(`-- Add your SQL code here:
+SELECT 1;
+
+-- Define SQL parameters in your definition file:
+-- SELECT * from users where user_id = :user_id;
```
leeweisberger, line 25: Where are the parameters being defined? What's a definition file? Is this an example?

francescaleung: yea, this probably needs some work. There's not really an equivalent `console.log(parameters)` here, which is sort of what this is trying to be. Should it add an example of parameter information that would ideally go in the task def file?

leeweisberger: I see what you're trying to do I think.

SELECT 1 is the simplest example of sql.
SELECT * from users where user_id = :user_id; is the simplest example of referencing params that are declared in the defn file.

hmm... We could
- Link to docs
- Expect the user to reference docs
- Make this read more like docs:
```
To reference parameters, use `:parameter`. e.g. `SELECT * from users where user_id = :user_id`
```

I guess if you generate your first sql file, this is useful. If you are generating your 50th sql file, this is useless :). Maybe we should just lean on our docs instead of including documentation every time one generates a task?

francescaleung:
```sql
-- Add your SQL queries here. You can include multiple queries in this file.
-- See SQL documentation: https://docs.airplane.dev/creating-tasks/sql
SELECT 1;
```

Doesn't reference parameters at all, but again, there's no real `console.log(params)` equivalent.

leeweisberger (approved):

pkg/runtime/sql/sql.go
leeweisberger: I think it makes sense what you have... I do think we could delete the ":user_id" thing, but up to you

PR #36, opened 2022-01-24
Properly handle query / entrypoint contents in SQL definition

francescaleung:
This adds proper handling to the `query` property in KindOptions + SQL entrypoint contents. It also adds SetEntrypoint to Definition_0_3 & GetQuery to SQLDefinition_0_3.

Supports AIR-2461.

### DIFF
--- a/pkg/deploy/taskdir/definitions/def_0_3.go
+++ b/pkg/deploy/taskdir/definitions/def_0_3.go
@@ -407,6 +407,20 @@ type SQLDefinition_0_3 struct {
 	Resource   string                 `json:"resource"`
 	Entrypoint string                 `json:"entrypoint"`
 	Parameters map[string]interface{} `json:"parameters,omitempty"`
+
+	// Contents of Entrypoint, cached
+	entrypointContents string `json:"-"`
+}
+
+func (d *SQLDefinition_0_3) GetQuery() (string, error) {
+	if d.entrypointContents == "" {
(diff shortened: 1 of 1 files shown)


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1023,6 +1088,15 @@ func (d *Definition_0_3) GetSlug() string {
 	return d.Slug
 }
 
+func (d *Definition_0_3) SetEntrypoint(entrypoint string) error {
```
leeweisberger, line 1091: Can you explain what the point of this is? Looks like every kind implements this basically the same way except some throw an error. What's the usecase?

leeweisberger (approved):

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: stupid answer: so we can programmatically set the entrypoint outside of a def file / task struct.

longer answer: Sometimes the entrypoint isn't defined, even when an entrypoint is required. e.g., when a Python task is created in the web frontend, it doesn't get created with an entrypoint defined. When we run `airplane init`, the user is asked where the entrypoint should be, & if it's not being run with `--code-only`, a definition file is produced. That definition file needs to have the entrypoint. `SetEntrypoint` is a nice way of abstracting away the  task kind implementations.

PR #38, opened 2022-01-28
Add env parameter to API client

colinking:
This PR adds an `envSlug` parameter to the client API (`GetTask` and `UpdateTask`). This [will be used in the CLI](https://github.com/airplanedev/cli/pull/374) and (in a future PR) in the deployer.

cc @ameyakhare

### DIFF
--- a/pkg/deploy/discover/discover.go
+++ b/pkg/deploy/discover/discover.go
@@ -86,7 +86,9 @@ func (d *Discoverer) DiscoverTasks(ctx context.Context, paths ...string) ([]Task
 				// The file is not an Airplane task.
 				continue
 			}
-			task, err := d.Client.GetTask(ctx, slug)
+			task, err := d.Client.GetTask(ctx, api.GetTaskRequest{
+				Slug: slug,
+			})
 			if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/deploy/discover/discover.go
```diff
@@ -45,6 +45,12 @@ type Discoverer struct {
 	TaskDiscoverers []TaskDiscoverer
 	Client          api.IAPIClient
 	Logger          logger.Logger
+
+	// EnvSlug is the slug of the environment to look for discovered tasks in.
+	//
+	// If a task is discovered, but doesn't exist in this environment, then it will be ignored by
+	// default. Each TaskDiscover can optionally override this behavior.
```
francescaleung, line 52: is this behaviour TODO?

francescaleung (approved):

pkg/deploy/discover/discover.go
colinking: Yeah, not currently used.

colinking: Lemme just revert that for now, TBH. Will add back once we start on the deployer work.

PR #39, opened 2022-01-31
Pick interpolation mode when converting definition to UpdateTaskRequest

colinking:


colinking:
AD; doing this server-side instead.

PR #42, opened 2022-02-02
Copy execute rules over correctly

zhan-airplane:
Needs upgrading CLI after this too, the existing version will continue to wipe `executeRules` and I'm not sure if we can handle it correctly on the API side short of creating a new field entirely

---
cc @francescaleung @joshma

### DIFF
--- a/pkg/deploy/taskdir/definitions/def_0_3.go
+++ b/pkg/deploy/taskdir/definitions/def_0_3.go
@@ -872,6 +872,7 @@ func (d Definition_0_3) GetUpdateTaskRequest(ctx context.Context, client api.IAP
 	if currentTask != nil {
 		req.RequireExplicitPermissions = currentTask.RequireExplicitPermissions
 		req.Permissions = currentTask.Permissions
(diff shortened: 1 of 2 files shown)


leeweisberger (commented):
shouldn't we be fine with the pointer change you made on the API? We only update execute rules if it is provided. In fact this pr shouldn't even be needed right? Because we aren't updating execute rules?

zhan-airplane:
No, execute rules now is always provided, just that it's provided as setting everything to false (since right now if we're not setting it, it defaults to falsey values). 

I didn't make a pointer change to the API because of this - we need to do a pointer change on both the client and the server if we want to go with that route. My hotfix was instead to explicitly ignore `ExecuteRules` if the update task request was from the CLI, but we definitely want to be handling `ExecuteRules` at some point in the future, so it's not a long term solution.

Instead of just doing a pointer change on `ExecuteRules`, I think we should figure out what our golang client API should look like (e.g. should we put pointers on every single field in every `apiclient` struct?) and then do that change all in one go, and then tweak the hotfix I have on execute rules.

In that vein, I guess I'll close this PR first

PR #44, opened 2022-02-03
Support pip.conf when building python tasks

zhan-airplane:
If a `pip.conf` file is in the root, sets the `PIP_CONFIG_FILE` env var when running `pip install`. Feature request from https://app.intercom.com/a/apps/we6lj1ka/inbox/inbox/5188557/conversations/29392

---
cc @joshma


joshma:
Very cool, thanks for being proactive on this @zhan-airplane . Can you also update the Python docs?

leeweisberger (approved):
Seems fine. Did some reading and this makes sense. Can you include this in our documentation?

PR #45, opened 2022-02-09
Handle the same task being discovered by multiple discoverers

francescaleung:
Returns a single task config per slug, in order of discoverers, in order of discovery. E.g.: given discoverers [DefnDiscoverer, ScriptDiscoverer], if task `my_task` has both a linked script & a defn file, the defn file will take precedence; given two defn files for the same task, the first defn file to be discovered takes precedence.

Resolves AIR-2973.

### DIFF
--- a/pkg/deploy/discover/discover.go
+++ b/pkg/deploy/discover/discover.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"sort"
 
 	"github.com/airplanedev/lib/pkg/api"
 	"github.com/airplanedev/lib/pkg/deploy/taskdir/definitions"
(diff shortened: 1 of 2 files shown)


pkg/deploy/discover/discover.go
```diff
@@ -116,9 +126,44 @@ func (d *Discoverer) DiscoverTasks(ctx context.Context, paths ...string) ([]Task
...
+			}
+			for _, tc := range tcs {
+				if td.TaskConfigSource() == tc.From {
+					taskConfigs[i] = tc
+					found = true
+					break
+				}
+			}
+		}
+	}
```
leeweisberger, line 167: Can you break this out into a helper (at least one) and/or add a few comments? it makes sense, just requires me to really read every line to tell what is going on - unfortunately go is super verbose here :(

pkg/deploy/discover/discover.go
```diff
@@ -116,9 +126,44 @@ func (d *Discoverer) DiscoverTasks(ctx context.Context, paths ...string) ([]Task
...
+	sort.Strings(slugs)
+
+	taskConfigs := make([]TaskConfig, len(slugs))
+	for i, slug := range slugs {
+		tcs := taskConfigsBySlug[slug]
+		if len(tcs) == 1 {
+			taskConfigs[i] = tcs[0]
+			continue
+		}
```
leeweisberger, line 153: You should be able to remove this. Up to you

leeweisberger (approved):
Make sure we bump this in both cli and api - we are also using the discoverer in the api now!

pkg/deploy/discover/discover.go
francescaleung: ehh, I like it

PR #47, opened 2022-02-09
Adjust TaskEntrypoint to be relative to defn file

francescaleung:
Resolves AIR-2976.

### DIFF
--- a/pkg/deploy/discover/discover_test.go
+++ b/pkg/deploy/discover/discover_test.go
@@ -15,7 +15,6 @@ import (
 
 func TestDiscoverTasks(t *testing.T) {
 	fixturesPath, _ := filepath.Abs("./fixtures")
-	discoverPath, _ := filepath.Abs("./")
 	tests := []struct {
 		name          string
 		paths         []string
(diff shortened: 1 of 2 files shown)


pkg/deploy/discover/discover_test.go
```diff
@@ -110,15 +109,13 @@ func TestDiscoverTasks(t *testing.T) {
...
-					TaskRoot: fixturesPath,
-					// TODO adjust to be fixturesPath when entrypoint is relative to task defn
-					TaskEntrypoint: discoverPath + "/single_task.js",
+					TaskRoot:       fixturesPath,
+					TaskEntrypoint: fixturesPath + "/single_task.js",
 					Def: &definitions.Definition_0_3{
 						Name:        "sunt in tempor eu",
 						Slug:        "my_task",
 						Description: "ut dolor sit officia ea",
-						// TODO adjust to be absolute path
```
francescaleung, line 120: I'm not convinced this actually needs to happen--should this be that different than the defns from the script files?

leeweisberger: I think it's fine as long as we are handling this correctly upstream.

PR #48, opened 2022-02-11
Handle root, workdir determination in defn discoverer as in script discoverer

francescaleung:
Root & workdir determination matches now between script & defn discoverer. We only care about root + workdir when there's an entrypoint, as far as I can tell.

Also removes WorkingDirectory from TaskConfig, as it was not being used (it is being shown to the user, if it's different from the TaskRoot, but it does not actually appear to be used anywhere).

Resolves AIR-2551, AIR-2986.


pkg/deploy/discover/defn_discover.go
```diff
@@ -45,25 +46,42 @@ func (dd *DefnDiscoverer) GetTaskConfig(ctx context.Context, task api.Task, file
...
+
+		wd, err := r.Workdir(absEntrypoint)
 		if err != nil {
 			return TaskConfig{}, err
 		}
+		def.SetWorkdir(taskroot, wd)
 	}
 
-	return TaskConfig{
-		TaskRoot:       dir.DefinitionRootPath(),
```
leeweisberger, line 62: are we still using definitionRootPath?

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1035,6 +1045,17 @@ func (d *Definition_0_3) SetEntrypoint(entrypoint string) error {
 	return taskKind.setEntrypoint(entrypoint)
 }
 
+func (d *Definition_0_3) SetWorkdir(taskroot, workdir string) error {
+	// TODO: currently only a concept on Node - should be generalized to all builders.
+	if d.Node == nil {
+		return nil
+	}
+
+	d.Node.Workdir = strings.TrimPrefix(workdir, taskroot)
```
leeweisberger, line 1054: Can we add a test for this where it isn't an empty string?

pkg/deploy/discover/defn_discover.go
```diff
@@ -45,25 +46,42 @@ func (dd *DefnDiscoverer) GetTaskConfig(ctx context.Context, task api.Task, file
...
+	}
+
 	entrypoint, err := def.Entrypoint()
 	if err == definitions.ErrNoEntrypoint {
 		// nothing
 	} else if err != nil {
 		return TaskConfig{}, err
 	} else {
-		taskFilePath, err = filepath.Abs(filepath.Join(filepath.Dir(dir.DefinitionPath()), entrypoint))
+		absEntrypoint, err := filepath.Abs(filepath.Join(filepath.Dir(dir.DefinitionPath()), entrypoint))
```
leeweisberger, line 60: Can you add a comment here and/or split it up? Pretty hard to follow what's going on. We are joining the definition path to the entrypoint - so the entrypoint is stored relative to the definition path? Is it also supplied by the user relatative to the defintion path? is that obvious?

leeweisberger (approved):
One q / clarification. The rest reads fine

pkg/deploy/discover/defn_discover.go
francescaleung: Only for legacy YAML deploys right now.

francescaleung:
@leeweisberger rerequesting review for sanity check

pkg/deploy/discover/defn_discover.go
```diff
@@ -45,25 +46,52 @@ func (dd *DefnDiscoverer) GetTaskConfig(ctx context.Context, task api.Task, file
...
+		def.SetWorkdir(taskroot, wd)
+
+		// Entrypoint for builder needs to be relative to taskroot, not definition directory.
+		if defnDir != taskroot {
+			ep, err := filepath.Rel(taskroot, absEntrypoint)
+			if err != nil {
+				return TaskConfig{}, err
+			}
+			def.SetEntrypoint(ep)
+		}
```
francescaleung, line 91: This feels gross, & I think it's because there's the definition as the user has defined it, & there's the definition as what we need it to be. I think this is going to cause problems for AIR-2974.

PR #51, opened 2022-02-14
Break out build config as a separate concept

francescaleung:
Differentiate somewhat between "options that are part of the task definition" & "options that are part of the build config". This uses task KindOptions as a base, & merges it with BuildConfig that is tracked separately. It enables us to tweak build-related options without actually changing the task definition (e.g., workdir for Node, or rewriting the entrypoint to be relative to the task root).

Downstream services should call GetBuildConfig instead of GetKindAndOptions for build options.


pkg/deploy/discover/defn_discover.go
```diff
@@ -79,15 +79,17 @@ func (dd *DefnDiscoverer) GetTaskConfig(ctx context.Context, task api.Task, file
 		if err != nil {
 			return TaskConfig{}, err
 		}
-		def.SetWorkdir(taskroot, wd)
+		if err := def.SetWorkdir(taskroot, wd); err != nil {
+			return TaskConfig{}, err
+		}
```
leeweisberger, line 84: Would have expected a linter to catch this

pkg/build/types.go
```diff
@@ -2,6 +2,8 @@ package build
 
 type KindOptions map[string]interface{}
 
+type BuildConfig map[string]interface{}
```
leeweisberger, line 5: Can you add a comment of what this is, especially vs kind options?

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1172,6 +1165,28 @@ func (d *Definition_0_3) convertTaskKindFromTask(ctx context.Context, client api
 	}
 }
 
+func (d *Definition_0_3) GetBuildConfig() (build.BuildConfig, error) {
+	_, options, err := d.GetKindAndOptions()
```
leeweisberger, line 1169: Kind of confused by this - why does the getter for build config rely on reading in kind options?

leeweisberger (changes requested):
I want to make sure I understand the building of build config from kind options. 

Is it because build config is a superset of kind options?

pkg/deploy/taskdir/definitions/def_0_3.go
leeweisberger: I re-read your pr description. I see the reasoning now.

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1172,6 +1165,28 @@ func (d *Definition_0_3) convertTaskKindFromTask(ctx context.Context, client api
...
+	if err != nil {
+		return nil, err
+	}
+	for k, v := range d.buildConfig {
+		if v == nil {
+			delete(options, k)
+		} else {
+			options[k] = v
+		}
+	}
```
leeweisberger, line 1179: I see what you're doing here now - this will work, but I think it shows intention clearer if you invert this.

Start with buildConfig being an empty map, and then add options to it if they don't already exist. Make sense?

leeweisberger:
> I want to make sure I understand the building of build config from kind options.
> 
> Is it because build config is a superset of kind options?

Edit - I see the reasoning - let me know what you think about the change to invert the build config creation logic to make it clearer what is happening

leeweisberger (approved):

PR #52, opened 2022-02-15
Use printf instead of echo

leeweisberger:
https://linear.app/airplane/issue/AIR-2922/[l3av]-shell-builder-breaks-on-custom-shell-zcli
Printf is standard across systems while echo is not:
image

In the case of zcli - printf respects newlines while echo does not.

---

There may be unintended side effects here, but I think we'll be fine....


joshma:
@colinking can you review as well? nice find @leeweisberger 

generally thinking I think we should be more defined on our shell builder "api" - e.g. it has to be an ubuntu base, you can only customize X Y Z - but this fix looks good

pkg/build/builder.go
```diff
@@ -353,7 +353,7 @@ func inlineString(s string) string {
 	// Since the string is wrapped in single-quotes, escape any single-quotes
 	// inside of the target string.
 	s = strings.ReplaceAll(s, "'", `'"'"'`)
-	return "echo '" + s + "'"
+	return "printf '" + s + "'"
```
colinking, line 356: `printf` does not print a trailing new line, unlike `echo`. Double check if we need it?

colinking (approved):

pkg/build/builder.go
leeweisberger: I went through all of our usecases of this and don't think we need it.

PR #53, opened 2022-02-16
Remove Workdir from NodeDef03

francescaleung:
Got lost somewhere in merges & reshuffling. Workdir is now being handled through build config, not kind options.


leeweisberger (approved):
:( I have definitely seen this already haha

PR #55, opened 2022-02-16
Use Def03/DefinitionInterface everywhere

francescaleung:
Upgrading lib with this patch effectively releases Def03 & deprecates Def02.

* Removes Definition alias. Concrete instances satisfying the DefinitionInterface explicitly need to instantiate a specific Definition version. Prefer to specify DefinitionInterface as return values or parameter values.
* Moves Definition functions to be under Def02, since the Definition alias is removed.
* Removes Root from Def03. It wasn't being used, it's not part of the schema. Along with this, removes DefinitionRootPath from the TaskDirectory.

For upstream clients:
* ScriptDiscoverer now needs an IAPIClient.

---

cc @joshma


pkg/deploy/taskdir/definitions/def_0_2.go
```diff
@@ -87,6 +93,301 @@ type RESTDefinition struct {
 	FormDataBody       map[string]string `yaml:"formDataBody,omitempty" mapstructure:"formDataBody,omitempty"`
 }
 
-func (d Definition_0_2) upgrade() (Definition, error) {
-	return Definition(d), nil
+func NewDefinitionFromTask_0_2(task api.Task) (Definition_0_2, error) {
```
francescaleung, line 96: Most of this is moved from definitions.go, & modified for Definition -> Definition_0_2. Some renaming (e.g., this function used to be called NewDefinitionFromTask).

PR #57, opened 2022-02-17
Fix js dev - find package.json in workdir, not root

leeweisberger:
Look for a package.json in the working directory rather than the root. 

Fixes https://linear.app/airplane/issue/AIR-2710/support-packagejson-not-in-root-folder


colinking (approved):
🔥

PR #60, opened 2022-02-22
Clean up schema & Def03

francescaleung:
Resolves AIR-3066, AIR-3067.


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1004,11 +947,7 @@ func (d Definition_0_3) Entrypoint() (string, error) {
 }
 
 func (d *Definition_0_3) UpgradeJST() error {
-	taskKind, err := d.taskKind()
-	if err != nil {
-		return err
-	}
-	return taskKind.upgradeJST()
+	return nil
```
leeweisberger, line 950: 1. I'd prefer if we just delete `UpgradeJST` if we aren't supporting it.
2. Would rather throw an error than return nil so that we don't have someone silently call this thinking they are doing something.

leeweisberger (approved):

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: It's still doing something for Def02. It's only Def03 where it's a no-op.

PR #62, opened 2022-02-23
Fix SQL entrypoint to be relative to defn file, not working directory

francescaleung:
Adds notion of an absolute entrypoint to the definition, so that the SQL definition can properly find its entrypoint.

Resolves AIR-3063.


leeweisberger (commented):
Can you explain why this only applies to sql? How does this work for node e.g.? Isn't the entrypoint also relative to the definition path?

Also, why can't we store the entrypoint as an absolute path

francescaleung:
I think it's a little much to expect user-defined definition files to have absolute paths. The entrypoint should be relative to where the defn file is. (This is on my list of "important things to document".)

This is specific to SQL because it's the only one where we actually need to read the entrypoint to fill in some of the task definition (specifically, the `query` in KindOptions). In every other definition, the entrypoint is akin to metadata: important, but we don't actually need to know anything about it. All other entrypoint-reading happens outside of the definition; here is the only place where it happens _inside_.

leeweisberger:
> I think it's a little much to expect user-defined definition files to have absolute paths. The entrypoint should be relative to where the defn file is. (This is on my list of "important things to document".)

By "why can't we store the absolute path", I meant the user can supply the path relative and then we can store it absolute to the defn file so we don't have to store the defn path. i.e. the only reason you're storing the defn path is to get at the entrypoint, so why not store the entrypoint as an absolute path (there might be something preventing us from doing this - but what is it??)

francescaleung:
ahh, I see. Yes, we can store the absolute entrypoint, though I'd prefer we store it separately from the user-provided entrypoint. In theory this probably works about the same, but maybe it'll feel better. I'll try it out.

pkg/deploy/taskdir/definitions/def_0_2.go
```diff
@@ -252,6 +252,10 @@ func (def *Definition_0_2) SetWorkdir(taskroot, workdir string) error {
 	return nil
 }
 
+func (def *Definition_0_2) SetDefinitionPath(path string) error {
+	return nil
+}
+
```
leeweisberger, line 258: revert?

pkg/deploy/taskdir/definitions/interface.go
```diff
@@ -18,18 +18,30 @@ type DefinitionInterface interface {
...
 	GetKindAndOptions() (build.TaskKind, build.KindOptions, error)
 	GetEnv() (api.TaskEnv, error)
 	GetSlug() string
 	UpgradeJST() error
 	GetUpdateTaskRequest(ctx context.Context, client api.IAPIClient, currentTask *api.Task) (api.UpdateTaskRequest, error)
 	SetWorkdir(taskroot, workdir string) error
 
-	// Entrypoint returns ErrEntrypoint if the definition doesn't define an entrypoint. May be empty.
+	// Entrypoint returns ErrNoEntrypoint if the task kind definition requires no entrypoint. May be
+	// empty. This is relative to the defn file, if one exists; otherwise, it's not super useful.
```
leeweisberger, line 39: It is relative to the defn file if the user specifies it that way right?  But technically they don't have to do that? Might want to adjust the comment

leeweisberger (approved):

PR #64, opened 2022-02-23
Add SupportsLocalExecution to Runtime interface

francescaleung:
Shortcut to if `PrepareRun` returns `ErrNotImplemented`.

Supports AIR-3060, AIR-3061.


pkg/runtime/runtime.go
```diff
@@ -78,6 +78,10 @@ type Interface interface {
 	// If running the script locally is not supported the method returns
 	// an `ErrNotImplemented`.
 	PrepareRun(ctx context.Context, logger logger.Logger, opts PrepareRunOptions) (rexprs []string, closer io.Closer, err error)
+
+	// SupportsLocalExecution returns true if PrepareRun does not return
+	// `ErrNotImplemented`.
```
leeweisberger, line 83: This is kind of an implementation detail. Maybe just say what this does - returns true if the runtime can be ran locally

leeweisberger (approved):

PR #65, opened 2022-02-23
Rename SQL parameters -> query args

francescaleung:
Resolves AIR-2521.

### DIFF
--- a/pkg/deploy/taskdir/definitions/def_0_3.go
+++ b/pkg/deploy/taskdir/definitions/def_0_3.go
@@ -395,7 +395,7 @@ var _ taskKind_0_3 = &SQLDefinition_0_3{}
 type SQLDefinition_0_3 struct {
 	Resource   string                 `json:"resource"`
 	Entrypoint string                 `json:"entrypoint"`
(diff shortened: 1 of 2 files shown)


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -395,7 +395,7 @@ var _ taskKind_0_3 = &SQLDefinition_0_3{}
 type SQLDefinition_0_3 struct {
 	Resource   string                 `json:"resource"`
 	Entrypoint string                 `json:"entrypoint"`
-	Parameters map[string]interface{} `json:"parameters,omitempty"`
+	QueryArgs  map[string]interface{} `json:"queryArgs,omitempty"`
```
colinking, line 398: Do we set query args when initializing a SQL task w/ `airplane init`? Just curious if we need to update that codepath, too.

colinking (approved):

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: We don't, no.

PR #68, opened 2022-02-25
Add comments to generated YAML file

francescaleung:
There needs to be another pass for copy, but this pass is for the generation part. It's somewhat ugly & the end result is less than ideal, but short of switching libraries (again!) this might be the best we could do.

Resolves AIR-3174.

### DIFF
--- a/pkg/deploy/taskdir/definitions/def_0_3.go
+++ b/pkg/deploy/taskdir/definitions/def_0_3.go
@@ -1,6 +1,7 @@
 package definitions
 
 import (
+	"bytes"
 	"context"
 	_ "embed"
 	"encoding/json"
@@ -18,8 +19,8 @@ import (
 )
 
 type Definition_0_3 struct {
-	Name        string                    `json:"name"`
 	Slug        string                    `json:"slug"`
+	Name        string                    `json:"name"`
 	Description string                    `json:"description,omitempty"`
 	Parameters  []ParameterDefinition_0_3 `json:"parameters,omitempty"`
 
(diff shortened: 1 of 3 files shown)


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -42,6 +44,171 @@ type Definition_0_3 struct {
...
+// getCommentMap dynamically creates the comment map for the definition. This is necessary because
+// the yaml library we're using runs into nil pointer exceptions if you try to leave a comment on a
+// field that doesn't end up in the YAML file (e.g., because it's empty). Returns the comment map
+// plus a set of bytes that should be tacked on to the end of the YAML file (because it's
+// impossible to specify comments to go at the end of the YAML file in the CommentMap).
+func (d Definition_0_3) getCommentMap() (yaml.CommentMap, []byte, error) {
+	cm := yaml.CommentMap{}
+	nextComments := []string{}
+	addComment := func(c comment) {
+		// Add an empty line, then break the text into lines of max 78 characters.
```
leeweisberger, line 63: So we need a newline before each comment? Does it look like
```
#
# mycomment
```

That's pretty crummy...

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -42,6 +44,171 @@ type Definition_0_3 struct {
...
+				nextComments = append(nextComments, " "+e)
+			}
+		}
+	}
+
+	addComment(comment{
+		key:   "$.slug",
+		doAdd: true,
+		text:  "Used by Airplane to identify your task. Do not change.",
+	})
```
leeweisberger, line 84: Can we create a yaml file with these comments? One place where we can update them?

Or better yet it would be great if we could just pull these from the json schema - is that something we would do? We have the text and example in the json schema and we just run through that and create comments from it.

Might be tough with the doAdd stuff since that is dynamic... what do you think the best way to maintain this is?

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -42,6 +44,171 @@ type Definition_0_3 struct {
...
+	example string
+	text    string
+}
+
+// getCommentMap dynamically creates the comment map for the definition. This is necessary because
+// the yaml library we're using runs into nil pointer exceptions if you try to leave a comment on a
+// field that doesn't end up in the YAML file (e.g., because it's empty). Returns the comment map
+// plus a set of bytes that should be tacked on to the end of the YAML file (because it's
+// impossible to specify comments to go at the end of the YAML file in the CommentMap).
+func (d Definition_0_3) getCommentMap() (yaml.CommentMap, []byte, error) {
```
leeweisberger, line 59: Is there a way we can test this? It's gonna be kind of tough to make changes here and not break things... I want a golden file of what we expect the defn to look like with comments and then ensure that we produce exactly that.

leeweisberger (changes requested):
Would like to hear some thoughts on a better way to maintain this than adding new function calls

And tests please

francescaleung:
Superseded by https://github.com/airplanedev/lib/pull/77.

PR #70, opened 2022-02-25
Add EnvVars to task <> definition conversion

francescaleung:
Somehow this got missed.

Resolves AIR-3175.


leeweisberger (approved):
Would be nice if we had a test that would have caught this

PR #75, opened 2022-03-01
Support deploying tasks to new environments

colinking:
To support deploying tasks into environments where they are not already deployed, we need to update our deployment logic to not depend on a full `Task` struct.

First off, I removed the `Task` struct from the `TaskDiscoverer` interface's `GetTaskConfig` method. Only script discovery needed to do a full `GetTask` call, so I've moved that into the `GetTaskConfig` method instead.

Second, I removed the `Task` struct from the `Definition` interface's `GetUpdateTaskRequest` method. We used the task to populate a subset of fields with the task's current value. Instead, I've moved these fields to pointer types (see API PR below).

Depends on:
- https://github.com/airplanedev/airport/pull/1737
- https://github.com/airplanedev/airport/pull/1740

Used by:
- https://github.com/airplanedev/cli/pull/412
- https://github.com/airplanedev/airport/pull/1735

cc @ameyakhare


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -809,12 +810,16 @@ func (d Definition_0_3) taskKind() (taskKind_0_3, error) {
...
-func (d Definition_0_3) GetUpdateTaskRequest(ctx context.Context, client api.IAPIClient, currentTask *api.Task) (api.UpdateTaskRequest, error) {
+func (d Definition_0_3) GetUpdateTaskRequest(ctx context.Context, client api.IAPIClient) (api.UpdateTaskRequest, error) {
 	req := api.UpdateTaskRequest{
 		Slug:        d.Slug,
 		Name:        d.Name,
 		Description: d.Description,
 		Timeout:     d.Timeout,
+		ExecuteRules: api.UpdateExecuteRulesRequest{
+			RequireRequests:     &d.RequireRequests,
+			DisallowSelfApprove: pointers.Bool(false),
```
francescaleung, line 821: I'd rather drop this, default nil seems safer

pkg/deploy/taskdir/definitions/interface.go
```diff
@@ -20,6 +20,7 @@ type DefinitionInterface interface {
 	GetKindAndOptions() (build.TaskKind, build.KindOptions, error)
 	GetEnv() (api.TaskEnv, error)
 	GetSlug() string
+	GetName() string
```
francescaleung, line 23: What's this for?

pkg/runtime/comment.go
```diff
@@ -27,39 +27,36 @@ func Comment(r Interface, taskURL string) string {
 	return r.FormatComment("Linked to " + taskURL + " [do not edit this line]")
 }
 
-// Slug returns the slug from the given file.
-//
-// Ok is true if the slug was found and isn't empty.
-func Slug(filePath string) (slug string, ok bool) {
+// Slug returns the slug from the given file. An empty string is returned if a slug was not found.
+func Slug(filePath string) string {
```
francescaleung, line 31: I rather liked the explicit `ok` return value, but up to you

pkg/utils/pointers/pointers.go
```diff
@@ -0,0 +1,13 @@
+package pointers
```
francescaleung, line 1: yea, this is probably overdue

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -825,22 +830,15 @@ func (d Definition_0_3) GetUpdateTaskRequest(ctx context.Context, client api.IAP
 		req.Constraints = *d.Constraints
 	}
 
-	if d.RequireRequests {
-		req.ExecuteRules.RequireRequests = true
-	}
-	if d.AllowSelfApprovals != nil && !*d.AllowSelfApprovals {
-		req.ExecuteRules.DisallowSelfApprove = true
+	if d.AllowSelfApprovals != nil {
+		disallow := !*d.AllowSelfApprovals
```
francescaleung, line 834: maybe we need a pointers.ToBool, would make this easier to read

pkg/deploy/discover/script_discover.go
```diff
@@ -16,57 +16,75 @@ import (
...
+		EnvSlug: sd.EnvSlug,
+	})
 	if err != nil {
-		return TaskConfig{}, errors.Wrapf(err, "cannot determine how to deploy %q - check your CLI is up to date", file)
+		var merr *api.TaskMissingError
+		if !errors.As(err, &merr) {
+			return nil, errors.Wrap(err, "unable to get task")
+		}
+
+		return nil, nil
```
francescaleung, line 46: So we're not logging that we've found a valid file but it won't deploy because the task doesn't exist? if the task isn't in the environment, do we still fall into this case?

pkg/deploy/discover/defn_discover.go
```diff
@@ -9,119 +9,137 @@ import (
...
+		return nil, err
 	}
 
 	tc := TaskConfig{
-		Task: task,
-		Def:  &def,
+		Def:    &def,
+		Source: dd.TaskConfigSource(),
+		// By definition, all task definitions use JSTs:
+		TaskInterpolationMode: "jst",
```
francescaleung, line 69: False! Not guaranteed. True in 99% of cases, probably.

francescaleung (changes requested):
Mostly okay, but the interpolation assumption for definition files is incorrect.

Also I think @leeweisberger should get looped in for the discoverer changes.

pkg/deploy/taskdir/definitions/def_0_3.go
colinking: So if you have this in a task YAML:

```yaml
allowSelfApprove: false
```

And then you remove that and re-deploy, wouldn't you expect that to clear to some default value?

pkg/deploy/taskdir/definitions/interface.go
colinking: https://github.com/airplanedev/cli/pull/412/files#diff-c80696cdcbfd2853cf9e9c0054a1bee04de602ac383c0b678af62504dc13b90fR186

pkg/deploy/taskdir/definitions/def_0_3.go
colinking: Maybe a `pointers.NegateBool(d.AllowSelfApprovals)`? A `ToBool` on its own wouldn't help much here. Going to punt on this.

pkg/deploy/discover/script_discover.go
colinking: Good catch. I didn't mean to remove the log. I'll bring that back.

If a task doesn't exist in an environment, we would currently not deploy it. I'll open a ticket to discuss supporting that (e.g. we could copy the config from another environment).

colinking: @francescaleung brought the log message back!

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: Ahh okay, I'd rather this bit get looped into the AllowSelfApprovals bit on line 833 then.

francescaleung (approved):

pkg/deploy/discover/script_discover.go
```diff
@@ -12,61 +12,81 @@ import (
...
+func (sd *ScriptDiscoverer) GetTaskConfig(ctx context.Context, file string) (*TaskConfig, error) {
+	slug := runtime.Slug(file)
+	if slug == "" {
+		return nil, nil
+	}
+
+	task, err := sd.Client.GetTask(ctx, api.GetTaskRequest{
+		Slug:    slug,
+		EnvSlug: sd.EnvSlug,
+	})
```
francescaleung, line 41: If a task doesn't exist in this environment, does that mean you can't deploy-via-script?

leeweisberger:
I walked through this, but there's way too much going on into this pr to keep track of everything. I can't really confidently review this. Up to @francescaleung whether she's confident in the review here

pkg/deploy/discover/script_discover.go
colinking: For now, no. There's a ticket for that: https://linear.app/airplane/issue/AIR-3223/deploy-code-only-tasks-to-new-environments

PR #77, opened 2022-03-02
Add GenerateCommentedFile to (sometimes) generate a commented YAML file

francescaleung:
Resolves AIR-3174, AIR-2490.

---
cc @joshma

### DIFF
--- a/pkg/deploy/taskdir/definitions/def_0_3.go
+++ b/pkg/deploy/taskdir/definitions/def_0_3.go
@@ -1,11 +1,13 @@
 package definitions
 
 import (
+	"bytes"
 	"context"
 	_ "embed"
 	"encoding/json"
 	"os"
 	"strings"
+	"text/template"
 
 	"github.com/airplanedev/lib/pkg/api"
 	"github.com/airplanedev/lib/pkg/build"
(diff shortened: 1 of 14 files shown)


joshma:
maybe add tests that the generated yaml is valid yaml?

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -821,6 +823,116 @@ func (d Definition_0_3) Marshal(format TaskDefFormat) ([]byte, error) {
...
+func (d Definition_0_3) GenerateCommentedFile(format TaskDefFormat) ([]byte, error) {
+	// If it's not YAML, or you have other things defined on your task def, bail.
+	if format != TaskDefFormatYAML ||
+		d.Description != "" ||
+		len(d.Parameters) > 0 ||
+		len(d.Constraints) > 0 ||
+		d.RequireRequests ||
+		(d.AllowSelfApprovals != nil && !*d.AllowSelfApprovals) ||
+		d.Timeout > 0 {
+		return d.Marshal(format)
```
leeweisberger, line 838: Rather than calculate this, shouldn't we know whether a task yaml already has stuff when we call this function? We should have a path where we are creating a defn from scratch and one where we are creating it from a pre-existing task. Can we pass a bool into this function with whether the task has already been initialized?

pkg/deploy/taskdir/definitions/yaml_templates.go
```diff
@@ -0,0 +1,197 @@
+package definitions
+
+const definitionTemplate = `
```
leeweisberger, line 3: Can we add tests that fill in templated values and then validate that these are valid? I think that's pretty important before merging.

leeweisberger (changes requested):
Approach looks good. Question about determing whether a yaml file is new. Also need tests before approving

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: You're suggesting that the caller should determine whether the commented YAML can be generated? That seems more dicey than this calculation, & this calculation is already pretty dicey... I can have a bool somewhere indicating whether the def was initialized via NewDefinition or NewDefinitionFromTask, but since all fields are public anyway, callers technically have the ability to tweak things outside of Definition methods. It wouldn't be ideal, but it's technically possible.

leeweisberger: Don't we know at some point whether we have a previously existing task or this is brand new? And isn't that the only information we need?

leeweisberger: I guess the task might be previously existing but have none of this information, in which case we want to add the comments... so I think I'm wrong

leeweisberger (approved):

PR #81, opened 2022-03-07
Respect working directory of custom Dockerfile

leeweisberger:
If the user specifies a custom Dockerfile when running a shell task, run the shell script in the same directory as the custom Dockerfile. Currently, we copy the shell script into /airplane - if a user wants to access files created by their custom Dockerfile, it isn't obvious that they are now inside the /airplane directory. This is more intuitive - our shim no longer changes the working directory and it's as if the shell script is run at the end of the custom Dockerfile.

https://linear.app/airplane/issue/AIR-2842/shell-builder-should-work-with-different-workdir


pkg/build/shell.go
```diff
@@ -22,12 +22,14 @@ func shell(root string, options KindOptions) (string, error) {
...
 	// Build off of the dockerfile if provided:
 	var dockerfileTemplate string
+	var workDir string
 	if dockerfilePath := FindDockerfile(root); dockerfilePath != "" {
 		contents, err := ioutil.ReadFile(dockerfilePath)
 		if err != nil {
 			return "", errors.Wrap(err, "opening dockerfile")
 		}
 		dockerfileTemplate = string(contents)
+		workDir = "."
```
colinking, line 32: LGTM. This is a breaking change. Can we identify users that are relying on the prior `/airplane` behavior?

pkg/examples/shell/diffworkdir/main.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Linked to https://app.airplane.dev/t/new_shell_task [do not edit this line]
+# Params are in environment variables as PARAM_{SLUG}, e.g. PARAM_USER_ID
+
+# test.sh is created in the custom Dockerfile in directory /myScripts.
+# Testing this ensures that we execute the main shell script within the same
+# working directory as the custom Dockerfile.
+./test.sh
```
colinking, line 8: 🔥

colinking, line 8: 🔥

colinking (approved):

pkg/build/shell.go
colinking: We should also document this working directory behavior in our shell docs: https://docs.airplane.dev/creating-tasks/shell#customize-the-environment-with-dockerfile

leeweisberger: It is a breaking change and I'll add docs.

I don't think there's any way for us to identify users that are relying on the previous behavior - we would need to search through the contents of their dockerfiles and look for users using a custom WorkDir in their custom dockerfile. Even if we could do that, we couldn't confidently identify whether people are relying on old behavior.

Options:
1. Try to add logging that detects a backwards incompatible issue. The easiest way would probably be to log if someone is using a WORKDIR in their custom dockerfile.
2. Live with the breaking change - push the breaking change, bump the minor version of the CLI by 1, and include the breaking change in the release notes. Update the docs. Tasks will break if users update their CLI and try to run their shell scripts. The surface area of breakages here is most likely very low - zero

I think 2 is the best option, but want to run by you first.

colinking: Sounds good with me.

One remaining concern is that these breaking changes will get picked up by our GitHub action since it uses the `:latest` tag: https://github.com/airplanedev/airplane-deploy/blob/baf528e305be07c07bbe705d14ca601b0bc744e0/Dockerfile#L1

PR #83, opened 2022-03-08
Update test workflow and run on all pushes

colinking:
Noticed that tests were not running on stacked PRs. While fixing that, I updated our test workflow to match airport.


francescaleung (approved):
:tada: my hero

PR #84, opened 2022-03-08
Rename SuggestExt -> SuggestExts, return multiple applicable extensions

francescaleung:
Supports AIR-3224.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"sort"
 
 	"github.com/airplanedev/lib/pkg/build"
 	"github.com/airplanedev/lib/pkg/utils/logger"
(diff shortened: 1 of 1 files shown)


pkg/runtime/runtime.go
```diff
@@ -137,14 +138,17 @@ func Lookup(path string, kind build.TaskKind) (Interface, error) {
...
+	exts := []string{}
 	for ext, runtime := range runtimes {
 		if runtime.Kind() == kind {
-			return ext
+			exts = append(exts, ext)
 		}
 	}
-	return ""
+	// Sort, so the return value is deterministic.
+	sort.Strings(exts)
```
leeweisberger, line 150: We should be opinionated on which ext we show first. e.g. do we want to show .js or .ts as the default (I'm not sure what the answer is, you might want to ask in #eng).

leeweisberger (changes requested):

leeweisberger (approved):

PR #90, opened 2022-03-17
Switch to python 3.9.6 bullseye

leeweisberger:
Upgrade python from buster -> (more up to date) bullseye. Use the closest version of python 3.9.4 -> 3.9.6.

This image has OpenSSL 1.1.1k  25 Mar 2021 instead of a 2019 version. 

FIXES AIR-2903

---

We determined that openssl is probably not causing any issues, but this upgrades us to a newer version.

We're skipping from d->k https://www.openssl.org/news/openssl-1.1.1-notes.html

### DIFF
--- a/pkg/build/versions.json
+++ b/pkg/build/versions.json
@@ -38,8 +38,8 @@
   "python": {
     "3": {
       "image": "registry.hub.docker.com/library/python",
-      "tag": "3.9.4-buster",
-      "digest": "sha256:b004a71e38f8ace26e7554d5c2fa802a8bb39a5818cbe10ab49fd0b408a40c20"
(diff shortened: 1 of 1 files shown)


zhan-airplane (commented):
Hmm looks like the unit test is failing...?

leeweisberger:
> Hmm looks like the unit test is failing...?

I copied the wrong sha. passing now

zhan-airplane (approved):
Would we need to update documentation anywhere?

joshma:
> Would we need to update documentation anywhere?

good q. not right now; we only document python version. i think we should document the OS specifics at some point, but it's sort of nice not to have to. maybe we never have to if the abstraction is "function"? 😇

PR #91, opened 2022-03-18
Use `yaml.MarshalWithOptions` with UseLiteralStyleIfMultiline option

francescaleung:
## Description

Before:
```
rest:
  resource: httpbin
  method: POST
  path: /post
  bodyType: json
  body: "{\n  \"name\": \"foo\",\n  \"number\": 30\n}\n"
```

After:
```
rest:
  resource: httpbin
  method: POST
  path: /post
  bodyType: json
  body: |
    {
      "name": "foo",
      "number": 30
    }
```

## Test plan

Run `airplane init` on a REST task with a string body that has newlines.

Resolves AIR-3315.


leeweisberger (approved):
beautiful. Would be nice if there was a test

PR #94, opened 2022-03-21
Handle parameters with config var options

francescaleung:
## Description

Updates schema + parameter handling for config var options. Resolves AIR-3351.

## Test plan

Run `airplane init` from a task with a parameter with config var options. Run `airplane deploy` on resulting definition file.

### DIFF
--- a/pkg/deploy/taskdir/definitions/def_0_3.go
+++ b/pkg/deploy/taskdir/definitions/def_0_3.go
@@ -6,6 +6,7 @@ import (
 	_ "embed"
 	"encoding/json"
 	"os"
+	"reflect"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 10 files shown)


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1290,9 +1299,34 @@ func (d *Definition_0_3) convertParametersFromTask(ctx context.Context, client a
...
+							Config: &configNameStr,
+						}
+					}
+				case reflect.String, reflect.Bool, reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
+					reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64,
+					reflect.Float32, reflect.Float64:
+					p.Options[j] = OptionDefinition_0_3{
+						Label: opt.Label,
+						Value: opt.Value,
+					}
```
leeweisberger, line 1327: What are you trying to solve with this? Why can't this just be the default?

pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -341,7 +341,7 @@
           "type": "boolean"
         },
         "options": {
-          "description": "A list of options to constrain the parameter values.",
+          "description": "A list of options to constrain the parameter values. Can be a single value, an object with a label and a value, or an object with a label and a config var.",
```
leeweisberger, line 344: We should update our reference docs with an example here

leeweisberger (approved):

PR #98, opened 2022-03-24
Add TransactionMode to SQL tasks

francescaleung:
## Description

New SQL tasks get created with a default TransactionMode of auto. Initializing SQL tasks without a TransactionMode results in no transactionMode entry in the definition file. Resolves AIR-2896.

## Test plan

Deploy a SQL task with a transaction mode. Initialize a SQL task from an existing task with a transaction mode, and without a transaction mode.

### DIFF
--- a/pkg/deploy/taskdir/definitions/def_0_3.go
+++ b/pkg/deploy/taskdir/definitions/def_0_3.go
@@ -797,7 +797,8 @@ func NewDefinition_0_3(name string, slug string, kind build.TaskKind, entrypoint
 		}
 	case build.TaskKindSQL:
 		def.SQL = &SQLDefinition_0_3{
-			Entrypoint: entrypoint,
+			Entrypoint:      entrypoint,
+			TransactionMode: "auto",
 		}
 	case build.TaskKindREST:
 		def.REST = &RESTDefinition_0_3{
--- a/pkg/deploy/taskdir/definitions/def_0_3.go
+++ b/pkg/deploy/taskdir/definitions/def_0_3.go
@@ -467,11 +467,26 @@ type SQLDefinition_0_3 struct {
 	Resource        string                 `json:"resource"`
 	Entrypoint      string                 `json:"entrypoint"`
 	QueryArgs       map[string]interface{} `json:"queryArgs,omitempty"`
-	TransactionMode string                 `json:"transactionMode,omitempty"`
+	TransactionMode SQLTransactionMode     `json:"transactionMode,omitempty"`
 
 	// Contents of Entrypoint, cached
 	entrypointContents string `json:"-"`
 	absoluteEntrypoint string `json:"-"`
+}
+
+type SQLTransactionMode string
+
+var _ yaml.IsZeroer = SQLTransactionMode("")
+
+func (tm SQLTransactionMode) IsZero() bool {
+	return tm == "auto" || tm == ""
+}
+
+func (tm SQLTransactionMode) Value() string {
+	if tm == "" {
+		return "auto"
+	}
+	return string(tm)
 }
 
 func (d *SQLDefinition_0_3) GetQuery() (string, error) {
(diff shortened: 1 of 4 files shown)


francescaleung:
Blocked on AIR-3395.

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -785,7 +797,8 @@ func NewDefinition_0_3(name string, slug string, kind build.TaskKind, entrypoint
 		}
 	case build.TaskKindSQL:
 		def.SQL = &SQLDefinition_0_3{
-			Entrypoint: entrypoint,
+			Entrypoint:      entrypoint,
+			TransactionMode: "auto",
```
colinking, line 801: What do you think of not writing this field if `auto` when initializing a SQL task? And then we treat "not set" as "auto".

Ideally these YAML files are as minimal as possible.

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -561,11 +569,15 @@ func (d *SQLDefinition_0_3) getKindOptions() (build.KindOptions, error) {
...
-	return build.KindOptions{
+	options := build.KindOptions{
 		"entrypoint": d.Entrypoint,
 		"query":      query,
 		"queryArgs":  d.QueryArgs,
-	}, nil
+	}
+	if d.TransactionMode != "" {
+		options["transactionMode"] = d.TransactionMode
+	}
```
colinking, line 579: If so, then we'd default this to `auto` here if not set.

colinking: And then the API would fallback on the task's current value (else auto) if it gets no value for the transaction mode.

pkg/deploy/taskdir/definitions/fixtures/sql.task.yaml
```diff
@@ -52,6 +52,10 @@ sql:
   # queryArgs:
   #   name: "{{params.name}}"
 
+  # The transaction mode to use. Valid values: auto, readOnly, readWrite, none.
+  # Default: auto.
+  transactionMode: auto
```
colinking, line 57: In this case, we would also comment this out like the other optional fields:

```suggestion
  # transactionMode: auto
```

pkg/deploy/taskdir/definitions/yaml_templates.go
```diff
@@ -156,6 +156,10 @@ sql:
   # (https://docs.airplane.dev/runbooks/javascript-templates).
   # queryArgs:
   #   name: "{{"{{params.name}}"}}"
+
+  # The transaction mode to use. Valid values: auto, readOnly, readWrite, none.
+  # Default: auto.
+  transactionMode: {{.TransactionMode}}
```
colinking, line 162: And this would be gated behind whether `.TransactionMode` is non-empty.

colinking (changes requested):
Have some thoughts on how we handle default values here. Can chat more over Slack if that's faster.

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -464,15 +464,31 @@ func (d *ShellDefinition_0_3) getEnv() (api.TaskEnv, error) {
...
+	TransactionMode SQLTransactionMode     `json:"transactionMode,omitempty"`
 
 	// Contents of Entrypoint, cached
 	entrypointContents string `json:"-"`
 	absoluteEntrypoint string `json:"-"`
 }
 
+type SQLTransactionMode string
+
+var _ yaml.IsZeroer = SQLTransactionMode("")
```
colinking, line 479: Checking IIUC; this logic is only needed for initializing a YAML from a task, since we init a `SQLDefinition_0_3` then serialize it as YAML?

colinking (approved):

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: This logic is needed to tell the yaml serializer that `TransactionMode=auto` should not be written out. In practice this is only for existing tasks that are initialized; for brand new tasks, we use our yaml template instead.

PR #104, opened 2022-04-12
Add configs to task definition yaml

calufornia:
## Description
This PR adds a `Configs` field to the task definition YAML for REST and SQL tasks, similar to how there is an `EnvVars` field for code tasks. It also adds a `Configs` field to a couple structs so that the CLI can deploy a task with config attachments. 

## Test plan
Unit tests

### DIFF
--- a/pkg/api/client.go
+++ b/pkg/api/client.go
@@ -20,28 +20,28 @@ type IAPIClient interface {
 
 // Task represents a task.
 type Task struct {
-	URL                        string             `json:"-" yaml:"-"`
-	ID                         string             `json:"taskID" yaml:"id"`
-	Name                       string             `json:"name" yaml:"name"`
-	Slug                       string             `json:"slug" yaml:"slug"`
-	Description                string             `json:"description" yaml:"description"`
-	Image                      *string            `json:"image" yaml:"image"`
-	Command                    []string           `json:"command" yaml:"command"`
-	Arguments                  []string           `json:"arguments" yaml:"arguments"`
-	Parameters                 Parameters         `json:"parameters" yaml:"parameters"`
-	Configs                    []ConfigAttachment `json:"configs" yaml:"configs"`
-	Constraints                RunConstraints     `json:"constraints" yaml:"constraints"`
-	Env                        TaskEnv            `json:"env" yaml:"env"`
-	ResourceRequests           ResourceRequests   `json:"resourceRequests" yaml:"resourceRequests"`
-	Resources                  Resources          `json:"resources" yaml:"resources"`
-	Kind                       build.TaskKind     `json:"kind" yaml:"kind"`
-	KindOptions                build.KindOptions  `json:"kindOptions" yaml:"kindOptions"`
-	Repo                       string             `json:"repo" yaml:"repo"`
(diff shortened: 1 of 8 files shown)


pkg/deploy/taskdir/definitions/fixtures/rest.task.yaml
```diff
@@ -72,6 +72,11 @@ rest:
   # formData:
   #   name: Alfred Pennyworth
 
+  # A list of config variables that are attached to this task.
+  # configs:
+  #   - config_variable_1
```
leeweisberger, line 77: Can you use a more realistic example here?

pkg/deploy/taskdir/definitions/fixtures/rest.task.yaml
```diff
@@ -72,6 +72,11 @@ rest:
   # formData:
   #   name: Alfred Pennyworth
 
+  # A list of config variables that are attached to this task.
```
leeweisberger, line 75: I think "attached to this task" is kind of confusing. Maybe "that this task can access"?

pkg/api/client.go
```diff
@@ -76,6 +77,7 @@ type UpdateTaskRequest struct {
 	Command                    []string                  `json:"command"`
 	Arguments                  []string                  `json:"arguments"`
 	Parameters                 Parameters                `json:"parameters"`
+	Configs                    []ConfigAttachment        `json:"configs"`
```
leeweisberger, line 80: I missed this in your last pr - []ConfigAttachment should probably be a pointer so that it uses a patch semantic. Without this, an old version of the CLI will delete config attachements because this code will not be in yet.

leeweisberger (changes requested):

pkg/api/client.go
```diff
@@ -77,7 +77,7 @@ type UpdateTaskRequest struct {
 	Command                    []string                  `json:"command"`
 	Arguments                  []string                  `json:"arguments"`
 	Parameters                 Parameters                `json:"parameters"`
-	Configs                    []ConfigAttachment        `json:"configs"`
+	Configs                    *[]ConfigAttachment       `json:"configs"`
```
leeweisberger, line 80: So go ahead and make this update in the api before this is merged

pkg/api/client.go
```diff
@@ -20,28 +20,28 @@ type IAPIClient interface {
...
+	URL                        string              `json:"-" yaml:"-"`
+	ID                         string              `json:"taskID" yaml:"id"`
+	Name                       string              `json:"name" yaml:"name"`
+	Slug                       string              `json:"slug" yaml:"slug"`
+	Description                string              `json:"description" yaml:"description"`
+	Image                      *string             `json:"image" yaml:"image"`
+	Command                    []string            `json:"command" yaml:"command"`
+	Arguments                  []string            `json:"arguments" yaml:"arguments"`
+	Parameters                 Parameters          `json:"parameters" yaml:"parameters"`
+	Configs                    *[]ConfigAttachment `json:"configs" yaml:"configs"`
```
leeweisberger, line 32: This can stay as not a pointer

leeweisberger (approved):

PR #106, opened 2022-04-14
Add max output line option to output parsing

zhan-airplane:
## Description
Part of better OOM handling - agents will use this to filter out output lines which are too long.

## Test plan
Unit tests

### DIFF
--- a/pkg/outputs/parse.go
+++ b/pkg/outputs/parse.go
@@ -23,6 +23,8 @@ type ParsedLine struct {
 	Value    ojson.Value
 }
 
+var ErrOutputLineTooLong = errors.New("output line too long")
+
 func parseOutputValue(matches []string) ojson.Value {
 	vs := strings.TrimSpace(matches[4])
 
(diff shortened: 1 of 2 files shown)


pkg/outputs/parse.go
```diff
@@ -147,12 +149,21 @@ func parseLogLineWithChunks(text string, chunks map[string]*strings.Builder) (st
 	return text, nil
 }
 
-func Parse(chunks map[string]*strings.Builder, logText string) (*ParsedLine, error) {
+type ParseOptions struct {
+	// If non-zero, returns an error if an output line is too long.
+	MaxOutputLineLen int
```
colinking, line 154: Let's specify the unit / how this value is used here:

```suggestion
	// OutputLineMaxBytes is the maximum number of bytes a single line of output (across
	// all output chunks) can contain. Any lines larger than this will return an error.
	//
	// Disabled if <= 0.
	OutputLineMaxBytes int
```

pkg/outputs/parse.go
```diff
@@ -147,12 +149,21 @@ func parseLogLineWithChunks(text string, chunks map[string]*strings.Builder) (st
...
+}
+
+func Parse(chunks map[string]*strings.Builder, logText string, opts ParseOptions) (*ParsedLine, error) {
 	outputText, err := parseLogLineWithChunks(logText, chunks)
 	if err != nil {
 		return nil, err
 	}
 
+	if opts.MaxOutputLineLen != 0 && opts.MaxOutputLineLen < len(outputText) {
```
colinking, line 163: Minor: handle negative values as "disabled"

```suggestion
	if opts.MaxOutputLineLen > 0 && opts.MaxOutputLineLen < len(outputText) {
```

colinking (approved):

PR #108, opened 2022-04-19
Add defn file path to DefinitionInterface

leeweisberger:
We want to know the absolute path to the defn file for git deploys.


pkg/deploy/discover/discover_test.go
```diff
@@ -157,6 +158,7 @@ func TestDiscoverTasks(t *testing.T) {
 					"workdir": "",
 				},
 			},
+			defnFilePath: "fixtures/defn.task.yaml",
```
francescaleung, line 161: these aren't absolute; should they be?

francescaleung (changes requested):

pkg/deploy/discover/discover_test.go
leeweisberger: Woops yeah. Updated

francescaleung (approved):

PR #109, opened 2022-04-19
Discover tasks by script if exists a defn with the same name

leeweisberger:
## Description
If a user runs `airplane deploy my_task.ts` and there is a file `my_task.task.yaml` in the same directory, we will deploy `my_task.task.yaml`. This only works if the task script is in the same directory as the task yaml.

## Test plan
Unit test. Plus I will test using the CLI when this is merged

FIXES AIR-3325


pkg/deploy/discover/discover.go
```diff
@@ -35,9 +35,6 @@ type TaskConfig struct {
 }
 
 type TaskDiscoverer interface {
-	// IsAirplaneTask inspects a file and if that file represents an Airplane task, it returns
-	// that task's slug. If that file is not a task, it will return an empty string.
-	IsAirplaneTask(ctx context.Context, file string) (string, error)
```
leeweisberger, line 40: this was dead code

francescaleung (approved):
Does deploying a directory get weird? Do you just have multiple identical task configs at the end & that's okay because we do a dedupe pass & they're identical anyway?

Can you add a test case for deploying a directory with yaml + script?

leeweisberger:
> Does deploying a directory get weird? Do you just have multiple identical task configs at the end & that's okay because we do a dedupe pass & they're identical anyway?

Exactly. This should be basically the same behavior as when you deploy a directory with a linked script and a task defn. We discover both and dedupe. I added a test for the specific case you called out

PR #110, opened 2022-04-19
Add full task def reference to starter task

leeweisberger:
Add link to full task def reference to starter task. Rather than linking to individual languages, I decided to link to the full page b/c many of the fields are under "Standard Fields".

FIXES AIR-3570


francescaleung (changes requested):
you updated the tests, I don't think you updated the template though

leeweisberger:
> you updated the tests, I don't think you updated the template though

at least the tests failed!

francescaleung (approved):

PR #111, opened 2022-04-20
Add build args to Dockerfile for node and python

leeweisberger:
## Description
In order to use build args in a Dockerfile, the arg must be declared in the Dockerfile. We are currently passing all user env vars as build args, but not declaring these args in the Dockerfile, so they cannot be used.

This will require one more Airport PR where we pass build arg keys when generating the Dockerfiles.

## Test plan
Wrote tests for python and node. Will test manually once the Airport PR is out

AIR-3569


pkg/build/builder.go
```diff
@@ -63,14 +63,15 @@ type LocalConfig struct {
 	// If nil, Push will produce an error.
 	Auth *RegistryAuth
 
-	// BuildEnv is a map of build-time environment variables to use.
-	BuildEnv map[string]string
+	// BuildArgs is a map of build-time environment variables to use.
+	BuildArgs map[string]string
```
leeweisberger, line 67: I renamed this to map to use the same name docker uses

zhan-airplane (approved):

PR #114, opened 2022-04-21
Resolve issues with importing latest Node SDK beta

colinking:
## Description

1. Resolves an issue where users could not install their own version of the `airplane` SDK (e.g. the new `0.2.0` beta). Resolves: https://linear.app/airplane/issue/AIR-3585/users-cannot-override-node-shim-dependencies
2. Resolves an issue where the `airplane` SDK, which depends on `node-fetch` starting in `0.2.0`, was getting marked as `external` and therefore not converted to CommonJS.

Due to (1), this is a breaking change for any Node tasks that (1) depend on a `0.0.x` version of the Node SDK and (2) use the output methods (`setOutput`, `appendOutput`) introduced in `0.1.x`. IOW, these tasks are implicitly relying on this bug to give them the newer version of the SDK. The fix is to update your `package.json` to reference the version of the SDK you want, e.g. `~0.1.2`. Many of our examples had this issue and I went through and made that change. I'm considering adding a safeguard, where we check if the user is requesting a `0.0.x` version of `airplane` and if so, continue to install the newer shim version of airplane. However, (1) we would need to parse node semver syntax (probably easier to just regex it?) and (2) as we bump the version of the SDK that the shim uses, this could cause issues for these customers. Because of (2) and because this only impacts rebuilds, I lean towards one-off helping customers who hit this issue.

Resolves AIR-3585

## Test plan

Ran locally and confirmed the built image runs successfully. Tested with a few variants of JS tasks (including w/ and w/out a dep on `airplane`).


colinking:
* main
  * **PR #114** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/114" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/114?utm_source=stack-comment).

pkg/build/node_externals.go
```diff
@@ -8,6 +8,12 @@ import (
 	"github.com/pkg/errors"
 )
 
+var esmModules = []string{
+	"node-fetch",
+	// airplane>=0.2.0 depends on node-fetch
+	"airplane",
```
colinking, line 14: We should replace this with an esbuild plugin that traverses the full dependencies tree. Not bothering for now.

pkg/build/node.go
```diff
@@ -187,28 +187,41 @@ func node(root string, options KindOptions, buildArgs []string) (string, error)
 		{{if .PostInstallCommand}}
 		RUN {{.PostInstallCommand}}
 		{{end}}
-		
+
 		RUN {{.InlineShim}} > /airplane/.airplane/shim.js && \
 			esbuild /airplane/.airplane/shim.js \
 				--bundle \
-				--external:airplane \
```
colinking, line 194: It's not clear to me why we did this. @leeweisberger do you recall?

colinking: [Looks like it was introduced here](https://github.com/airplanedev/cli/pull/284/files#diff-ec43be39c17f62e7efec68daefdf4c09d14912309d82c098160f72ced6244e0eR143) (cc @zhan-airplane)

leeweisberger: I would not think we want this

pkg/build/node.go
```diff
@@ -187,28 +187,41 @@ func node(root string, options KindOptions, buildArgs []string) (string, error)
...
+	if err != nil {
+		return nil, err
+	}
+
+	pjson := struct {
 		Dependencies map[string]string `json:"dependencies"`
 	}{
 		Dependencies: map[string]string{
 			"airplane":    "~0.1.2",
 			"@types/node": "^16",
```
leeweisberger, line 212: unrelated, but we shouldn't need this. This is just typescript types and we aren't even typechecking.

leeweisberger (approved):

pkg/build/node.go
leeweisberger: I'll handle this in another pr

PR #116, opened 2022-04-21
Stop installing typescript and types

leeweisberger:
We don't need to install typescript on our node images. We are immediately stripping types and converting to js when running esbuild.

We also don't need to install `@types/node` in our shim for the same reason.

Unfortunately this invalidates our docker build cache, but it will end up producing slimmer, faster images


colinking (approved):
LGTM, but give this a test with `airplane dev` too

PR #117, opened 2022-04-26
Discover apps in addition to tasks

leeweisberger:
## Description
Update the discoverer to discover apps while it is discovering tasks.

An app can only be configured with a definition file that ends in `.app.yaml`.

## Test plan
I've tested this locally and it works e2e. Also have a unit test for discovering a simple app.


pkg/api/client.go
```diff
@@ -14,6 +14,7 @@ type IAPIClient interface {
 	GetTask(ctx context.Context, req GetTaskRequest) (res Task, err error)
 	// GetTaskMetadata fetches a task's metadata by slug. If the slug does not match a task, a *TaskMissingError is returned.
 	GetTaskMetadata(ctx context.Context, slug string) (res TaskMetadata, err error)
+	GetApp(ctx context.Context, req GetAppRequest) (res App, err error)
```
leeweisberger, line 17: This obviously will not work until I push this functionality on the server

pkg/deploy/discover/discover.go
```diff
@@ -159,3 +201,47 @@ func (d Discoverer) deduplicateTaskConfigs(taskConfigsBySlug map[string][]TaskCo
 	}
 	return taskConfigs
 }
+
+// Given a map of slug -> [app config, ...], returns a list of app configs unique by slug, sorted
+// by slug. App configs are chosen based on order of AppDiscoverers & order of discovery.
+func (d Discoverer) deduplicateAppConfigs(appConfigsBySlug map[string][]AppConfig) []AppConfig {
```
leeweisberger, line 207: this is pretty much copy pasted from `deduplicateTaskConfigs`. Would be a great candidate for generics when we get go 18 working :exclamation:

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -889,16 +889,16 @@ func NewDefinition_0_3(name string, slug string, kind build.TaskKind, entrypoint
 	return def, nil
 }
 
-func (d Definition_0_3) Marshal(format TaskDefFormat) ([]byte, error) {
+func (d Definition_0_3) Marshal(format DefFormat) ([]byte, error) {
```
leeweisberger, line 892: renamed from TaskDefFormat -> DefFormat so we can share this between tasks/apps

pkg/deploy/discover/discover_test.go
```diff
@@ -337,12 +339,29 @@ func TestDiscoverTasks(t *testing.T) {
 			},
 			expectedErr: true,
 		},
+		{
+			name:  "app defn",
```
stepzhou, line 343: Nit: rename test function name to `TestDiscover`

pkg/deploy/discover/discover.go
stepzhou: This entire PR is mostly copy pasta code actually. This PR on top of the airport PR is really making me feel like we're missing some "buildable" interface here.

pkg/deploy/discover/discover.go
```diff
@@ -45,8 +55,17 @@ type TaskDiscoverer interface {
...
+	// IsAirplaneApp inspects a file and if that file represents an Airplane app, it returns
+	// that app's definition. If that file is not an app, it will return an empty definition.
+	IsAirplaneApp(ctx context.Context, file string) (*AppConfig, error)
+	// AppConfigSource returns a unique identifier of this TaskDiscoverer.
+	AppConfigSource() AppConfigSource
+}
+
 type Discoverer struct {
 	TaskDiscoverers []TaskDiscoverer
+	AppDiscoverers  []AppDiscoverer
```
stepzhou, line 68: Is there any way to have `TaskDiscoverer` and `AppDiscoverer` implement the same generic discoverer interface? They (including TaskConfig and AppConfig) are almost the same thing

pkg/deploy/taskdir/definitions/definitions.go
```diff
@@ -21,28 +25,46 @@ func NewDefinitionFromTask(ctx context.Context, client api.IAPIClient, t api.Tas
...
-	TaskDefFormatYAML    TaskDefFormat = "yaml"
-	TaskDefFormatJSON    TaskDefFormat = "json"
+	DefFormatUnknown DefFormat = ""
+	DefFormatYAML    DefFormat = "yaml"
+	DefFormatJSON    DefFormat = "json"
 )
 
 func IsTaskDef(fn string) bool {
-	return GetTaskDefFormat(fn) != TaskDefFormatUnknown
+	return GetTaskDefFormat(fn) != DefFormatUnknown
```
stepzhou, line 37: Nit: refactor out and call a more generic `GetDefFormat(fn, YamlTaskDefExtensions, JSONTaskDefExtensions) != DefFormatUnknown`

stepzhou (changes requested):

pkg/deploy/discover/discover.go
leeweisberger: I could only do this if AppConfig and TaskConfig were combined into one, probably via an interface. In order to do this, all of the code downstream in the deploy code (both in cli and airport) would only have access to the returned interface. This interface would have to have all of the information needed by the downstream deploy code - currently a lot of task-specific things. So this interface would have to have fns for all of the task specific things and AppConfig would have to return nil for all of these. We could probably trim down on the number of task specific things needed for deploying, but currently there's a good amount of them.

I don't think that ^ is doable without a good amount of rework (pretty large refactor in 3 repos) and some interface functions that only apply to one or the other (task or app) which kind of defeats the purpose of the interface.

Another approach would be to make a bunch of these helper functions generic (e.g. deduplicateAppConfigs/deduplicateTaskConfigs) so that they work for both app configs and task configs as long as they satisfied a simple interface. I'd like to do that, but we aren't using go 1.18 so we don't have access to generics.

tl/dr I agree, but getting a common interface right now is a hefty refactor (I'd also probably want to wait until we add more fields to AppConfig) that we could put as a stretch goal for this project. Getting rid of some of the duplication with generics seems like an easier more incremental goal if we can get go 1.18 running.

Thoughts?

leeweisberger: Using go 1.18 I updated deduplicateConfigs to be generic so it works for all discoverers

stepzhou: > I don't think that ^ is doable without a good amount of rework (pretty large refactor in 3 repos) and some interface functions that only apply to one or the other (task or app) which kind of defeats the purpose of the interface.

If it's not straight-forward refactor, then it makes sense to me to defer this refactor. I'm more ok with how manual all this is since this is the first time we're adding a new buildable type. If we ever expect to add more buildable types though, we should probably reconsider the refactor

stepzhou (approved):

PR #118, opened 2022-04-27
Support building an app

leeweisberger:
## Description
Support building a vite app. A few things here are very POC and will have to be revisited.

For now, we only support vite and assume that the app has been bootstrapped with vite and has a `vite.config.ts`.

We build the app and store the build in the `/dist` folder. After the image is built, we will upload the contents of the `dist` folder to gcs to be served.

## Test plan
Wrote a unit test to ensure build succeeds.

FIXES AIR-3623


pkg/build/app.go
```diff
@@ -0,0 +1,47 @@
...
+		}
+		return "", err
+	}
+
+	cfg := struct {
+		Base           string
+		InstallCommand string
+		OutDir         string
+	}{
+		Base:           "node@sha256:d388b6e0648a0f56765260daa20ee4533225845ee1e4d48886c3d0f7aaaa6384",
```
stepzhou, line 29: What is this sha?

pkg/build/app.go
```diff
@@ -0,0 +1,47 @@
...
+		FROM {{.Base}} as builder
+		WORKDIR /airplane
+
+		COPY package*.json yarn.* /airplane/
+		RUN {{.InstallCommand}}
+
+		COPY . /airplane/
+		RUN yarn build --outDir {{.OutDir}}
+
+		FROM scratch
```
stepzhou, line 44: For my knowledge: why use scratch here?

stepzhou (approved):

pkg/build/app.go
leeweisberger: I added a comment. It's a minimal (empty) image where we can copy the build artifacts

PR #124, opened 2022-05-05
Remove node v12 and v15 references from task definition template

calufornia:
## Description
Per https://linear.app/airplane/issue/AIR-3536/deprecate-node-12-and-15, we're deprecating support for Node v12 and v15. This PR removes references to these versions from the task definition YAML templates.

## Test plan
Unit tests

### DIFF
--- a/pkg/build/node_test.go
+++ b/pkg/build/node_test.go
@@ -81,15 +81,6 @@ func TestNodeBuilder(t *testing.T) {
 			Options: KindOptions{
 				"shim":        "true",
 				"entrypoint":  "main.ts",
-				"nodeVersion": "12",
-			},
-		},
-		{
-			Root: "typescript/esnext",
-			Kind: TaskKindNode,
-			Options: KindOptions{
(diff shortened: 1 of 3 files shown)


calufornia:
@colinking there's references to v12 and v15 in [versions.json](https://github.com/airplanedev/lib/blob/main/pkg/build/versions.json) as well as the JSON schema - do we need to keep these for tasks that might still be using these versions of node?

colinking:
> @colinking there's references to v12 and v15 in [versions.json](https://github.com/airplanedev/lib/blob/main/pkg/build/versions.json?rgh-link-date=2022-05-05T07%3A00%3A18Z) as well as the JSON schema - do we need to keep these for tasks that might still be using these versions of node?

When we are ready to remove support from the platform, we can remove those too. That will need to wait until everyone has migrated off.

colinking (approved):
LGTM assuming a +1 in DMs about the migration path.

PR #125, opened 2022-05-05
Add support for Node v18

calufornia:
## Description

We plan on adding support for node v18, per https://linear.app/airplane/issue/AIR-3549/support-node-18. This PR adds the latest v18 buster image to `versions.json` so that we can start building images with it.

Resolves AIR-3549

## Test plan
n/a

### DIFF
--- a/pkg/build/versions.json
+++ b/pkg/build/versions.json
@@ -1,5 +1,10 @@
 {
   "node": {
+    "18": {
+      "image": "registry.hub.docker.com/library/node",
+      "tag": "18.1.0-buster",
+      "digest": "sha256:9b2fc7ff617cd4b6beb123d452b56ad5fd3849dfaefda152c8d4c0836ccb1a9a"
+    },
     "16": {
       "image": "registry.hub.docker.com/library/node",
       "tag": "16.2.0-buster",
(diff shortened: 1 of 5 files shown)


pkg/build/versions.json
```diff
@@ -1,5 +1,10 @@
 {
   "node": {
+    "18": {
+      "image": "registry.hub.docker.com/library/node",
+      "tag": "18.1.0-buster",
+      "digest": "sha256:9b2fc7ff617cd4b6beb123d452b56ad5fd3849dfaefda152c8d4c0836ccb1a9a"
```
colinking, line 6: This looks off -- is this the wrong arch? You'll want to use the digest for `linux/amd64` from here: https://hub.docker.com/layers/node/library/node/18.1.0-buster/images/sha256-3b88dc7e2379f5adb2e2893e0d4cf1774e260c176e51a6a8fc0cb358a2749a3f?context=explore

colinking (approved):
LGTM!

pkg/build/versions.json
calufornia: ah, that digest is for buster-slim, will fix

PR #127, opened 2022-05-06
Upgrade sourcegraph

leeweisberger:
Upgrade sourcegraph lsif gh action which fixes sourcegraph with go 1.18


ameyakhare (approved):
:smoge:

PR #128, opened 2022-05-11
Handle default values in configvar parameters

francescaleung:
## Description

Previously, `airplane init --from slug` would create an invalid definition file if the task contained a configvar parameter with a default value. This patch adds handling for default values in configvar parameters.

Example usage:

```
parameters:
- name: ConfigVar
  slug: configvar
  type: configvar
  default:
    config: database_url
  options:
  - label: ""
    config: database_url
  - label: ""
    config: db/dsn:local
```

Resolves AIR-3672.

## Test plan

Unit tests, tested locally.

### DIFF
--- a/pkg/deploy/taskdir/definitions/def_0_3.go
+++ b/pkg/deploy/taskdir/definitions/def_0_3.go
@@ -1166,10 +1166,9 @@ func (d Definition_0_3) addParametersToUpdateTaskRequest(ctx context.Context, re
 	req.Parameters = make([]api.Parameter, len(d.Parameters))
 	for i, pd := range d.Parameters {
 		param := api.Parameter{
-			Name:    pd.Name,
-			Slug:    pd.Slug,
-			Desc:    pd.Description,
-			Default: pd.Default,
+			Name: pd.Name,
+			Slug: pd.Slug,
+			Desc: pd.Description,
 		}
 
 		switch pd.Type {
@@ -1185,6 +1184,26 @@ func (d Definition_0_3) addParametersToUpdateTaskRequest(ctx context.Context, re
 			param.Type = api.Type(pd.Type)
 		default:
 			return errors.Errorf("unknown parameter type: %s", pd.Type)
+		}
+
+		switch reflect.ValueOf(pd.Default).Kind() {
+		case reflect.Map:
+			m, ok := pd.Default.(map[string]interface{})
+			if !ok {
+				return errors.Errorf("unhandled default: %v", pd.Default)
+			}
+			if configName, ok := m["config"]; !ok {
+				return errors.Errorf("unhandled default: %v", pd.Default)
+			} else {
+				param.Default = map[string]interface{}{
+					"__airplaneType": "configvar",
+					"name":           configName,
+				}
+			}
+		case reflect.String, reflect.Bool, reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
+			reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64,
+			reflect.Float32, reflect.Float64:
+			param.Default = pd.Default
 		}
 
 		if pd.Required != nil && !*pd.Required {
@@ -1380,7 +1399,6 @@ func (d *Definition_0_3) convertParametersFromTask(ctx context.Context, client a
 			Name:        param.Name,
 			Slug:        param.Slug,
 			Description: param.Desc,
-			Default:     param.Default,
 		}
 
 		switch param.Type {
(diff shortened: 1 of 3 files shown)


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1187,6 +1186,26 @@ func (d Definition_0_3) addParametersToUpdateTaskRequest(ctx context.Context, re
...
+				return errors.Errorf("unhandled default: %v", pd.Default)
+			} else {
+				param.Default = map[string]interface{}{
+					"__airplaneType": "configvar",
+					"name":           configName,
+				}
+			}
+		case reflect.String, reflect.Bool, reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
+			reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64,
+			reflect.Float32, reflect.Float64:
```
colinking, line 1205: If you're going to list all types, add a `default` that reports an error for unhandled types

pkg/deploy/taskdir/definitions/def_0_3_test.go
```diff
@@ -985,6 +1030,24 @@ func TestDefinitionToUpdateTaskRequest_0_3(t *testing.T) {
 						Type:  "shorttext",
 						Regex: "foo.*",
 					},
+					{
+						Name: "Config var",
+						Slug: "config_var",
+						Type: "configvar",
+						Default: map[string]interface{}{
+							"config": "API_KEY",
+						},
```
colinking, line 1039: Why a map with a `config` field instead of a singular string value?

```suggestion
						Default: "API_KEY",
```

My take:
1. We should use the `type` field as a hint that informs how to parse the default value.
2. Once we support object parameter types, how do we identify this is a config not a map with a `"config"` field?
3. There's a case for consistency since we already use a dedicated `config` field in (a) env vars and (b) option values. But, IMO we should merge `config+value` in (b) as it can use `type` as in (1). OTOH, (a) has a good reason as it can't use (1) and does need to distinguish.

colinking (changes requested):
Tests look good! 🔥

pkg/deploy/taskdir/definitions/def_0_3_test.go
colinking: Was there a specific reason we went with the `config` field instead of using `value`? Didn't see a discussion of that on the original PR: https://github.com/airplanedev/lib/pull/94

If not, let's go with strings instead of maps here. Let's merge the `config` field on select options in with `value` sometime soon.

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1196,6 +1195,35 @@ func (d Definition_0_3) addParametersToUpdateTaskRequest(ctx context.Context, re
...
+			if pd.Type == "configvar" {
+				param.Default = map[string]interface{}{
+					"__airplaneType": "configvar",
+					"name":           pd.Default,
+				}
+			} else {
+				param.Default = pd.Default
+			}
+		default:
+			return errors.Errorf("unhandled default value: %T", pd.Default)
```
colinking, line 1224: Bit more clear for end users:

```suggestion
			return errors.Errorf("unsupported type for default value: %T", pd.Default)
```

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1196,6 +1195,35 @@ func (d Definition_0_3) addParametersToUpdateTaskRequest(ctx context.Context, re
 			return errors.Errorf("unknown parameter type: %s", pd.Type)
 		}
 
+		switch reflect.ValueOf(pd.Default).Kind() {
+		case reflect.Map:
+			m, ok := pd.Default.(map[string]interface{})
+			if !ok {
+				return errors.Errorf("unhandled default: %v", pd.Default)
+			}
+			if configName, ok := m["config"]; !ok {
```
colinking, line 1204: Only do this if `pd.Type == "configvar"`.

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1196,6 +1195,35 @@ func (d Definition_0_3) addParametersToUpdateTaskRequest(ctx context.Context, re
 			return errors.Errorf("unknown parameter type: %s", pd.Type)
 		}
 
+		switch reflect.ValueOf(pd.Default).Kind() {
+		case reflect.Map:
+			m, ok := pd.Default.(map[string]interface{})
+			if !ok {
+				return errors.Errorf("unhandled default: %v", pd.Default)
```
colinking, line 1202: These `unhandled default` errors are going to be confusing for users. Can you clarify those error messages?

PR #131, opened 2022-05-17
Parse schedules from yaml/json task defns

ameyakhare:
Resolves AIR-3725

## Description
Parses schedules from task defn files and exposes them on DefinitionInterface to be used by the CLI and api when deploying.

## Test plan
def_0_3_test.go:TestDefinitionMarshal_0_3
def_0_3_test.go:TestDefinitionGetSchedules_0_3

### DIFF
--- a/pkg/api/client.go
+++ b/pkg/api/client.go
@@ -350,3 +350,10 @@ type GetAppRequest struct {
 	ID   string
 	Slug string
 }
+
+type Schedule struct {
+	Name        string                 `json:"name"`
+	Description string                 `json:"description"`
+	CronExpr    string                 `json:"cronExpr"`
+	ParamValues map[string]interface{} `json:"paramValues"`
+}
--- a/pkg/deploy/discover/discover.go
+++ b/pkg/deploy/discover/discover.go
@@ -26,21 +26,12 @@ const (
 	ConfigSourceDefn   ConfigSource = "defn"
 )
 
-type ScheduleConfig struct {
-	Name        string
-	Description string
-	Slug        string
-	CronExpr    string
-	Params      map[string]interface{}
-}
-
 type TaskConfig struct {
 	TaskID         string
 	TaskRoot       string
 	TaskEntrypoint string
 	Def            definitions.DefinitionInterface
 	Source         ConfigSource
-	Schedules      []ScheduleConfig
 }
 
 func (c TaskConfig) GetSource() ConfigSource {
--- a/pkg/deploy/discover/discover.go
+++ b/pkg/deploy/discover/discover.go
@@ -26,12 +26,21 @@ const (
 	ConfigSourceDefn   ConfigSource = "defn"
 )
 
+type ScheduleConfig struct {
+	Name        string
+	Description string
+	Slug        string
+	CronExpr    string
+	Params      map[string]interface{}
+}
+
 type TaskConfig struct {
 	TaskID         string
 	TaskRoot       string
 	TaskEntrypoint string
 	Def            definitions.DefinitionInterface
 	Source         ConfigSource
+	Schedules      []ScheduleConfig
 }
 
 func (c TaskConfig) GetSource() ConfigSource {
--- a/pkg/deploy/taskdir/definitions/schema_0_3.json
+++ b/pkg/deploy/taskdir/definitions/schema_0_3.json
@@ -488,14 +488,24 @@
           "description": "(Experimental) A map of schedules that are to be deployed with this task. The key corresponds to a unique schedule across deploys.",
           "type": "object",
           "patternProperties": {
-            ".*": {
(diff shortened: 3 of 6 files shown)


pkg/deploy/discover/discover.go
```diff
@@ -26,12 +26,21 @@ const (
 	ConfigSourceDefn   ConfigSource = "defn"
 )
 
+type ScheduleConfig struct {
```
francescaleung, line 29: How is this different from the `Schedule` struct?

pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -482,6 +483,21 @@
           "description": "Set the runtime used for this task.",
           "enum": ["", "durable"],
           "default": ""
+        },
+        "schedules": {
+          "description": "Mapping of slugs to schedules for the task.",
+          "type": "object",
```
francescaleung, line 489: you should definitely describe this object--add `properties`, add `required`, add `additionalProperties=false`

I think also that we wanted a shorthand for `run_at_midnight: 0 0 * * *`?

pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -482,6 +483,21 @@
           "description": "Set the runtime used for this task.",
           "enum": ["", "durable"],
           "default": ""
+        },
+        "schedules": {
+          "description": "Mapping of slugs to schedules for the task.",
```
francescaleung, line 488: This could be more descriptive. It's not _false_ per se, just not particularly useful from a user's perspective. I would also add something in here cautioning users that this is an experimental feature for now (since the schema gets pulled in to e.g. vscode).

francescaleung (changes requested):

pkg/api/client.go
```diff
@@ -350,3 +350,10 @@ type GetAppRequest struct {
 	ID   string
 	Slug string
 }
+
+type Schedule struct {
```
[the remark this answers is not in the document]
ameyakhare: we pass schedules to the backend via a map where the slug is the key

pkg/deploy/discover/discover.go
ameyakhare: yeah I think this was a remnant from implementation, removed

pkg/deploy/taskdir/definitions/schema_0_3.json
ameyakhare: added properties, gonna skip the shorthand for now (since parsing gets a little annoying as we have to interface the field and type switch)

pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -482,6 +483,36 @@
           "description": "Set the runtime used for this task.",
           "enum": ["", "durable"],
           "default": ""
+        },
+        "schedules": {
+          "description": "(Experimental) A map of schedules that are to be deployed with this task. The key corresponds to a unique schedule across deploys.",
+          "type": "object",
+          "patternProperties": {
+            ".*": {
```
francescaleung, line 491: this should _probably_ be a slug, so `[a-z0-9_]*`, or whatever our normal regex for slugs is.

pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -482,6 +483,36 @@
...
+              "type": "object",
+              "properties": {
+                "name": { "type": "string" },
```
francescaleung, line 494: would be nice to have descriptions for these eventually

pkg/api/client.go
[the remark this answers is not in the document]
francescaleung: sure, I think it's useful to add slug here anyway--handling schedules in isolation (& not as a map collection) means you might want to know what the slug is & it's irritating to have to pass it around separately when it's a property of the schedule. Also, when we fetch schedules from the API server, do they come as a list or as a map?

pkg/deploy/taskdir/definitions/schema_0_3.json
francescaleung: If you need some inspiration, `OptionDefinition_0_3` has shorthand handling.

francescaleung (approved):

pkg/deploy/taskdir/definitions/schema_0_3.json
ameyakhare: added descriptions

pkg/deploy/taskdir/definitions/schema_0_3.json
ameyakhare: I'm doing `^[a-z0-9_]{1,50}$`

PR #132, opened 2022-05-18
Fix npm/yarn workspaces deploy & dev

leeweisberger:
We currently have a few bugs in our implementation of npm/yarn workspaces

When deploying, we only gather dependencies to mark as external from the root package.json. This means that any packages declared in a workspaces's package.json (generally where you declare all of your dependencies) are ignored and not marked external. This causes any dependencies that cannot be bundled to break the deployment.

When devving, we gather dependencies from the workspace only and not the root package.json (opposite as deploying :confounded: ). Moreover, we consider any sibling workspaces as dependencies and mark them external which causes us not to build them - this causes dev to not work if a workspace depends on another workspace (it often does).

Instead, what we have to do is:
1. Gather deps from all package.jsons of all of the workspaces (including the root) to mark as external... but
2. Do NOT mark any imported, sibling workspaces as external so that they are bundled.

To accomplish 2, we use the command `yarn workspaces info` which prints out the individual workspaces in a project and ensure these are not marked external so they can be built.

A few comments:
1. We require that users have `yarn` so we can run `yarn workspaces info`. We will install this in our own bundler so that we have it for deployments. We could remove this requirement, but it would mean we have to do the work of `yarn workspaces info` which seems pretty annoying
2. I still want to discuss where we want to move in terms of bundling in general.

Please let me know if we want to discuss any of this!


pkg/build/node_externals.go
```diff
@@ -21,20 +27,54 @@ var esmModules = []string{
...
+
+		if usesWorkspaces {
+			// If we are in a npm/yarn workspace, we want to bundle all packages in the same
+			// workspaces so they are run through esbuild.
+			yarnWorkspacePackages, err := getYarnWorkspacePackages(pathPackageJSON)
+			if err != nil {
+				return nil, err
+			}
+			for _, p := range yarnWorkspacePackages {
+				esmModules = append(esmModules, p)
```
yolken-airplane, line 62: Given that `esmModules` is a golang package-level map, this could cause issues if `ExternalPackages` is called multiple times? Also, the name might not be appropriate if the internal workspace package doesn't actually use ESM. How about creating another, function-scoped map like `workspaceModules` or `internalModules` or something like that then checking both in the deps loop below.

colinking:
@leeweisberger: AD during the tasks standup, I'm going to send this by @ameyakhare as the primary reviewer. I'll additionally review, but at a higher-level: https://consulting.drmaciver.com/code-review-quick-fixes/#designated-second-reviewers

pkg/build/node_externals.go
```diff
@@ -21,20 +27,54 @@ var esmModules = []string{
 // using just pg causes esbuild to bundle pg which bundles pg-native, which errors.
 // TODO: replace this with a cleaner esbuild plugin that can mark dependencies as external:
 // https://github.com/evanw/esbuild/issues/619#issuecomment-751995294
-func ExternalPackages(pathPackageJSON string) ([]string, error) {
-	var deps []string
-
-	allDeps, err := ListDependencies(pathPackageJSON)
+func ExternalPackages(rootPackageJSON string) ([]string, error) {
```
yolken-airplane, line 30: What do you think about having a test specifically for this function? Our current tests cover that builds work end-to-end, but I think it might be nice to have a verification that this function is spitting out the expected results under a variety of different conditions.

yolken-airplane (commented):
Looks good at a high level, just added questions about a few small things.

pkg/build/node.go
```diff
@@ -53,8 +53,8 @@ func node(root string, options KindOptions, buildArgs []string) (string, error)
 	}
 
 	workdir, _ := options["workdir"].(string)
-	pathPackageJSON := filepath.Join(root, "package.json")
-	hasPackageJSON := fsx.AssertExistsAll(pathPackageJSON) == nil
+	rootPackageJSON := filepath.Join(root, "package.json")
```
colinking, line 56: Let's search for the highest `package.json` between the script file and the root directory.

For example, for [this script](https://github.com/airplanedev/lib/tree/main/scripts/cache) we wanted to store the `package.json` in `/scripts/cache/package.json`. Since we need to access a JSON file in `/pkg/build/versions.json`, the root is overriden to `/`. Therefore, we had to store it in `/package.json` instead. IIRC, you fixed this bug but we just never got around to moving the `package.json`?

pkg/build/node_externals.go
```diff
@@ -85,3 +125,86 @@ func contains(list []string, needle string) bool {
...
+		return false, nil
+	}
+	var pkg pkgJSON
+	buf, err := os.ReadFile(pathPackageJSON)
+	if err != nil {
+		return false, errors.Wrapf(err, "node: reading %s", pathPackageJSON)
+	}
+
+	if err := json.Unmarshal(buf, &pkg); err != nil {
+		return false, fmt.Errorf("node: parsing %s - %w", pathPackageJSON, err)
```
colinking, line 142: nit: don't mix Go error wrapping w/ errors.Wrap (line 148). We use the latter.

pkg/build/node_externals.go
```diff
@@ -21,20 +27,54 @@ var esmModules = []string{
...
+	var deps []string
+	for _, pathPackageJSON := range pathPackageJSONs {
+		if pathPackageJSON == "" {
+			continue
+		}
+
+		if usesWorkspaces {
+			// If we are in a npm/yarn workspace, we want to bundle all packages in the same
+			// workspaces so they are run through esbuild.
+			yarnWorkspacePackages, err := getYarnWorkspacePackages(pathPackageJSON)
```
colinking: Interesting, so `yarn workspaces` can handle parsing the `npm` format for workspaces? Can you document that somewhere? 

Nit but if it handles both, rename this to `getWorkspacePackages`

colinking (commented):
Just skimming. Will defer to @ameyakhare

pkg/build/node_externals.go
leeweisberger: Added a test suite. I also found a bug where if you import a local dependency that is of a different version, we want to mark as external because npm/yarn will download this dep rather than use the local version

pkg/build/node_externals.go
leeweisberger: ya just a copy pasta

pkg/build/node_externals.go
leeweisberger: Yeah discussed this with Ben. As far as I can tell reading both of the docs they use the same format (very slightly nervous about this but I can't find any contradictions). And there is no npm equivalent, so it's either reuse yarn or roll our own which we could do if we have to

Updated comment and fn name

pkg/build/node_externals.go
```diff
@@ -76,12 +116,82 @@ func ListDependencies(pathPackageJSON string) ([]string, error) {
...
+// as yarn is installed.
+func getWorkspacePackages(pathPackageJSON string) (map[string]bool, error) {
+	cmd := exec.Command("yarn", "workspaces", "info")
+	cmd.Dir = filepath.Dir(pathPackageJSON)
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return nil, errors.Wrap(err, "reading yarn/npm workspaces. Do you have yarn installed?")
+	}
+	// out will be something like:
+	// yarn workspaces v1.22.17
```
ameyakhare, line 163: does the version of yarn affect the output in any way, or has that remained the same since this was introduced?

ameyakhare (approved):
overall looks fine to me, left some questions

pkg/build/node.go
leeweisberger: I am going to split this out into another PR. This still doesn't work as expected. This pr doesn't regress this at all, but it still has issues with using the correct package.json when the root is overriden.

I added this to my next sprint

PR #138, opened 2022-05-27
Rename schedule.params to schedule.paramValues

ameyakhare:
## Description
Adds consistency with how we name the field in the schedules UI.

## Test plan
👀


colinking (approved):
LGTM. Do we have any other references to `params` to update? docs, changelog, UI?

PR #139, opened 2022-05-27
Handle defaults in a more opinionated way

francescaleung:
## Description

This PR proposes a more opinionated way to handle (among other things) non-zero defaults. This can also be applied to places where there are multiple ways of expressing the same thing (e.g., config value options); this PR only addresses timeout (default=3600), & default true boolean properties (`parameters.required`, `allowSelfApprovals`).

The change to definition marshalling for JSON ensures that the custom `IsZero()` method gets used to determine if a field is zero/empty.

Resolves AIR-3698, AIR-3689.

## Test plan

Deployed definitions with explicit defaults & confirmed that the definition was reported as unchanged. Ensured that `airplane init --from <slug>` produced the same information for YAML vs JSON with regard to defaults.


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -795,14 +795,14 @@ func (d *RESTDefinition_0_3) getConfigAttachments() []api.ConfigAttachment {
...
-	Default     interface{}            `json:"default,omitempty"`
-	Required    *bool                  `json:"required,omitempty"`
-	Options     []OptionDefinition_0_3 `json:"options,omitempty"`
-	Regex       string                 `json:"regex,omitempty"`
+	Name        string                    `json:"name"`
+	Slug        string                    `json:"slug"`
+	Type        string                    `json:"type"`
+	Description string                    `json:"description,omitempty"`
+	Default     interface{}               `json:"default,omitempty"`
+	Required    DefaultTrueDefinition_0_3 `json:"required,omitempty"`
```
colinking, line 803: Do these need to be specific to `Definition_0_3`? Can we just have a `utils/defaults` package with `defaults.True` / `defaults.RunTimeout` / ... etc.?

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -902,18 +902,27 @@ func NewDefinition_0_3(name string, slug string, kind build.TaskKind, entrypoint
 func (d Definition_0_3) Marshal(format DefFormat) ([]byte, error) {
 	switch format {
 	case DefFormatYAML:
-		buf, err := yaml.MarshalWithOptions(d, yaml.UseLiteralStyleIfMultiline(true))
+		buf, err := yaml.MarshalWithOptions(d,
+			yaml.UseJSONMarshaler(),
```
colinking, line 906: Why does this use the JSON marshaler?  Is this b/c we implemented `MarhsalJSON/...`? Aren't there (iirc) MarshalYAML methods too?

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -902,18 +902,27 @@ func NewDefinition_0_3(name string, slug string, kind build.TaskKind, entrypoint
...
+			yaml.UseLiteralStyleIfMultiline(true))
 		if err != nil {
 			return nil, err
 		}
 		return buf, nil
 
 	case DefFormatJSON:
-		buf, err := json.MarshalIndent(d, "", "\t")
+		buf, err := yaml.MarshalWithOptions(d,
```
colinking, line 914: Why do we use the YAML marshaller here? Add a comment 🙏🏻

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1629,6 +1627,72 @@ func (d *Definition_0_3) SetBuildConfig(key string, value interface{}) {
...
+
+func (d DefaultTrueDefinition_0_3) IsZero() bool {
+	return d.Value() == true
+}
+
+type DefaultTimeoutDefinition_0_3 struct {
+	value int
+}
+
+var _ yaml.IsZeroer = &DefaultTimeoutDefinition_0_3{}
```
colinking, line 1671: What does having an IsZero method change?

colinking (approved):
Nice.

We may want to write our own marshaller and avoid using struct tags. I have some ideas if you want to chat sometime :)

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: Partially for tags, partially for MarshalJSON. The YAML library also deals with JSON; the built-in JSON library does not also deal with YAML. We probably should have leaned more heavily into YAML (tags + MarshalYAML). Hindsight is 20/20, I guess.

PR #140, opened 2022-05-31
Remove go, deno, dockerfile task types

leeweisberger:
Remove these deprecated task types. There are no go or deno tasks other than those owned by the airplane team. There is 1 dockerfile task, but it has never been run or updated and was created in march 2021.

FIXES AIR-2716


francescaleung (approved):
awww

PR #156, opened 2022-06-14
Update sample view to work with web app

leeweisberger:
Update sample view so that it can actually be deployed and render correctly 

image


rose-huang (approved):
🔥

PR #159, opened 2022-06-15
Monkey patch process.env in temporal workflow runtime

calufornia:
## Description

Adds an activity that loads in environment variables from the main node process. We then monkey patch `process.env` so that the user can continue to reference environment variables in their code in the same way as standard tasks. This `process.env` object will also contain some airplane-specific values that are used to determine which execute method in the SDK to call, env, etc.

## Test plan

Tested by deploying/running a workflow task locally without specifying `runtime: "workflow"` in the `ExecuteOptions`.


pkg/build/workflow/workflow-shim-activities.js
```diff
@@ -0,0 +1,14 @@
+// getEnvVars populates an object with envVarNames as the keys and values from the main Node thread's process.env.
+// We filter using envVarNames instead of doing a deep copy of the entire process.env because the majority of
+// environment variables do not need to be accessible by the user - only their custom env vars and the env vars required
+// by our SDKs should be present in the monkey patched process.env
+export async function getEnvVars(envVarNames) {
```
yolken-airplane, line 5: Cool. Out of curiosity- was this an approach recommended by Temporal or just something that you hacked together independently?

yolken-airplane (approved):

pkg/build/workflow/workflow-shim-activities.js
calufornia: Fetching env vars in an activity is actually recommended by Temporal (in some unpublished documentation): https://deploy-preview-1205--mystifying-fermi-1bc096.netlify.app/application-development-guide/#environment-variables - monkey patching `process.env` so that it's callable during workflow execution is separate 😄. Although there shouldn't be any reason why this wouldn't be recommended, since we're only modifying the global after we're already inside a workflow execution, and not injecting it into the vm, etc.

PR #162, opened 2022-06-16
Override base url with API url

leeweisberger:
## Description
To support proxying view content through the API, we need to point asset fetches at the API. This requires us to
1. Pass in the API host to hit at build time as an env var
2. Pass in the view id at build time as an env var so that the API knows which view we are getting content for
3. Update the base url to use these two pieces of data to construct a URL

The changes to the sample view are... just a sample. We're going to have to take this code and bake it into the vite config at build time.

## Test plan
This all manually works e2e!

### DIFF
--- a/pkg/build/app.go
+++ b/pkg/build/app.go
@@ -51,8 +51,8 @@ func app(root string, buildArgs []string) (string, error) {
 
 		{{.Args}}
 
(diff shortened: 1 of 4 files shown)


leeweisberger:
This is no longer necessary!

PR #170, opened 2022-07-01
Update workflow shim with renamed args

colinking:
## Description

Removed the shim activity for reading environment variables in favor of passing through all values. We don't set task revision environment variables since they may be JSTs. This also properly passes through other run env vars.

Resolves AIR-4155

cc @ameyakhare 

## Test plan

Ran locally and confirmed the workflow had access to resource attachment environment variables.


colinking:
* main
  * **PR #170** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/170" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/170?utm_source=stack-comment).

pkg/build/workflow/worker-and-activity-shim.js
```diff
@@ -2,9 +2,6 @@ import { NativeConnection, Worker } from '@temporalio/worker';
 
 // Activity code runs in the same node process as the worker, so we import it here directly.
 import { createActivities } from "airplane/internal/runtime/workflow"
-// TODO: Make this path configurable.
-import * as customActivities from "../activities"
```
colinking, line 6: Dropped customActivities since we don't plan to support it. One less change to make during local dev.

pkg/build/workflow/workflow-shim.js
```diff
@@ -3,22 +3,17 @@ import task from '{{.Entrypoint}}';
...
-export async function __airplaneEntrypoint(params, airplaneArgs) {
+export async function __airplaneEntrypoint(params, workflowArgs) {
   logger.info('airplane_status:started');
 
   try {
     // Monkey patch process.env
-    let env = await getEnvVars(airplaneArgs.EnvVarNames, airplaneArgs.RuntimeEnv);
     global.process = {
-      env
+      env: workflowArgs.EnvVars
```
colinking, line 16: Since we don't pass through task revision env vars, we can drop the activity entirely.

pkg/build/node.go
```diff
@@ -243,7 +241,6 @@ func node(root string, options KindOptions, buildArgs []string) (string, error)
 		RUN touch /airplane/activities.js
```
calufornia, line 241: Since you're dropping custom activities in this PR, let's also remove the logic here

calufornia (approved):

PR #171, opened 2022-07-05
support yarn 2

leeweisberger:
## Description
Support yarn 2 when building node tasks. 

`yarn workspace info` has been changed to `yarn workspace list` in newer versions of yarn. Detect what version of yarn the user is using, and then call the correct command and parse its output.

## Test plan
Added a unit test that failed before and passes with the new code path.

FIXES AIR-4029


pkg/build/node_externals.go
```diff
@@ -148,54 +149,103 @@ type workspaceInfoEntry struct {
 	WorkspaceDependencies []string `json:"workspaceDependencies"`
 }
 
-// getWorkspacePackages gets all local workspaces that are depended on by other workspaces.
-// It uses `yarn workspaces info` which does the heavy lifting of building out the dependency tree for us.
-// There is no npm workspaces equivalent, but `yarn workspaces info` works for both yarn and npm as long
-// as yarn is installed.
-func getWorkspacePackages(pathPackageJSON string) (map[string]bool, error) {
-	cmd := exec.Command("yarn", "workspaces", "info")
+func isYarn2(pathPackageJSON string) (bool, error) {
```
rose-huang, line 152: Rename to isYarnBerry and add a comment that Yarn Berry is yarn version 2+

rose-huang (approved):
🧶 🍓 
also lets update the docs at some point (ie  here https://docs.airplane.dev/creating-tasks/node-build and maybe in other places too)

PR #172, opened 2022-07-05
Fix TS type generation

leeweisberger:
## Description
When generating a typescript task from an existing task, we generate a type for the params. We had a bug in the code where we were no longer doing this.

Moreover, the type that we would have been generating was incorrect. I fixed the typing to be correct and extracted it into a reusable function.

## Test plan
Unit test to test the logic. I also generated a new ts task from an existing one and we now get a correct param type.

### DIFF
--- a/pkg/runtime/typescript/templates.go
+++ b/pkg/runtime/typescript/templates.go
@@ -1,0 +1,39 @@
+package typescript
+
+import "text/template"
+
+var code = template.Must(template.New("ts").Parse(`{{with .Comment -}}
+{{.}}
+
+{{end -}}
+{{.Params}}
+
+// Put the main logic of the task in this function.
+export default async function(params: Params) {
+  console.log('parameters:', params);
+
+  // You can return data to show outputs to users.
+  // Outputs documentation: https://docs.airplane.dev/tasks/outputs
+  return [
+    {element: 'hydrogen', weight: 1.008},
+    {element: 'helium', weight: 4.0026},
+  ];
+}
+`))
+
+type paramsTemplateConfig struct {
+	TaskName   string
+	TaskParams string
+}
+
+var paramsTemplate = template.Must(template.New("tsParams").Parse(`export type {{.TaskName}}Params = {
+  {{.TaskParams}}
+};
+`))
+
+var paramTypesTemplateNoParams = template.Must(template.New("tsParamsEmpty").Parse(`export type {{.TaskName}}Params = {};
+`))
+
+var fileType = `{ __airplaneType: "upload"; id: string; url: string }`
+
+var configType = `{ name: string; value: string }`
--- a/pkg/runtime/typescript/templates.go
+++ b/pkg/runtime/typescript/templates.go
@@ -7,7 +7,6 @@ var code = template.Must(template.New("ts").Parse(`{{with .Comment -}}
 
 {{end -}}
 {{.Params}}
-
 // Put the main logic of the task in this function.
 export default async function(params: Params) {
   console.log('parameters:', params);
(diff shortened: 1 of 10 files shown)


pkg/runtime/typescript/typescript.go
```diff
@@ -57,16 +29,22 @@ type Runtime struct {
 }
 
 // Generate implementation.
-func (r Runtime) Generate(t *runtime.Task) ([]byte, fs.FileMode, error) {
+func (r Runtime) Generate(t *runtime.Task, opts runtime.GenerateOpts) ([]byte, fs.FileMode, error) {
 	d := data{}
 	if t != nil {
-		d.Comment = runtime.Comment(r, t.URL)
+		if t.URL != "" && opts.GenerateComment {
```
leeweisberger, line 35: Now we are explicit about whether we want to generate a comment. We don't block type generation based on whether we want a comment

pkg/runtime/typescript/typescript.go
```diff
@@ -77,20 +55,57 @@ func (r Runtime) Generate(t *runtime.Task) ([]byte, fs.FileMode, error) {
...
+	var params []string
+	for slug, paramType := range parameters {
+		slug = slugToTypeKey(slug)
+		paramType := typeof(paramType)
+		params = append(params, fmt.Sprintf("%s: %s;", slug, paramType))
+	}
+
+	sort.Strings(params)
+	tc := paramsTemplateConfig{
+		TaskName:   strings.Title(strings.ReplaceAll(typePrefix, " ", "")),
```
colinking, line 69: Are you planning to use task prefix elsewhere?

pkg/runtime/typescript/typescript.go
```diff
@@ -77,20 +55,57 @@ func (r Runtime) Generate(t *runtime.Task) ([]byte, fs.FileMode, error) {
...
+
+	t := paramsTemplate
+	if len(params) == 0 {
+		t = paramTypesTemplateNoParams
+	}
+
+	var buff bytes.Buffer
+	err := t.Execute(&buff, tc)
+	if err != nil {
+		return "", err
```
colinking, line 81: Wrap the error

pkg/runtime/typescript/typescript.go
```diff
@@ -77,20 +55,57 @@ func (r Runtime) Generate(t *runtime.Task) ([]byte, fs.FileMode, error) {
...
+	case api.TypeConfigVar:
+		return configType
+	}
+	return "unknown"
+}
+
+// slugToTypeKey converts a slug to a key that can be used in a TypeScript type.
+func slugToTypeKey(slug string) string {
+	if strings.Contains(slug, "-") {
+		return fmt.Sprintf("\"%s\"", slug)
```
colinking, line 108: Slugs can't have dashes: https://github.com/airplanedev/airport/blob/1e9cf42688ae31a18d307ee732be504775117e84/pkg/utils/slugs.go#L20-L22

colinking: FWIW, it would be nice if params were provided as camelcase (rather than snakecase) to match JS conventions. Maybe our shim could one day translate automatically.

pkg/runtime/typescript/templates.go
```diff
@@ -0,0 +1,38 @@
+package typescript
+
+import "text/template"
```
colinking, line 3: nice

colinking (approved):
Good catch! 🔥

pkg/runtime/typescript/typescript.go
leeweisberger: Yeah it's in the gentypes pr

PR #176, opened 2022-07-07
Don't log during workflow replay

calufornia:
## Description

Sets `callDuringReplay` to `false` for the logger sink. This was previously required because [there was a bug](https://github.com/temporalio/sdk-typescript/issues/667) that caused the entire workflow (even steps that had not executed yet) to be flagged as replaying, which meant that any steps after a workflow interruption would not be logged if the field was not set to true. However, this introduced another bug - by enabling the flag, duplicate logs would occur for all workflow steps that occurred before the interrupt. Upgrading the airplane SDK version to one that uses the latest temporal typescript SDK lets us ensure that each log occurs exactly once.

Also contains various changes that are required as a result of breaking changes in the latest typescript SDK: https://github.com/temporalio/sdk-typescript/blob/main/CHANGELOG.md

cc @yolken-airplane 

## Test plan
Interrupted a workflow by delete the runner pods during a workflow execution, and verified that the task was still able to complete successfully.


pkg/build/workflow/worker-and-activity-shim.js
```diff
@@ -57,15 +57,18 @@ async function runWorker(params) {
             `airplane_workflow_log:workflow//${workflowInfo.workflowId}/${workflowInfo.runId} ${message}`
           );
         },
-        callDuringReplay: true,
+        callDuringReplay: false,
```
colinking, line 60: I'd like to get @yolken-airplane's review on this. My understanding is we reproduce logs during replay because our interceptors rely on logs for passing workflow metadata back to the runner. IIUC, disabling this would (?) cause issues. CMIIW!

pkg/build/workflow/workflow-bundler.js
```diff
@@ -1,16 +1,19 @@
 const worker = require('@temporalio/worker');
```
colinking, line 1: Guess we're going to need to handle breaking changes with multiple versions of the shim that we switch between, depending on what version of the Airplane SDK is installed 🤔

colinking (approved):
LGTM, but sanity check the replay change with @yolken-airplane if you haven't already. I'm curious what you learn!

pkg/build/workflow/worker-and-activity-shim.js
yolken-airplane: So I think with the SDK fix, this should do the right thing- logs (including ones for run state control) should be generated for events after the replay point. Let me know if it doesn't work as expected after deploying. :-)

pkg/build/workflow/workflow-bundler.js
calufornia: Yeah..the contributors of the typescript SDK did say they would be more careful about ensuring backwards compatibility from 1.0.0 forward, so hopefully don't need too many shim versions (if any).

PR #178, opened 2022-07-12
AIR-3982 Support failed workflows

kellymtrinh:
## Description
If a workflow fails (returns an error), we should log the error so the runbroker can interpret it and mark the run as failed. The previous behavior, which throws the error again, causes temporal to retry indefinitely.

## Test plan

Tested this locally with a runbroker change for both a task that has an activity and one that's only a workflow execution

Screen Shot 2022-07-11 at 4 28 25 PM

Screen Shot 2022-07-11 at 5 27 18 PM


calufornia (approved):
Nice!

PR #179, opened 2022-07-12
Patch extra files needed for views deployment

zhan-airplane:
## Description
Still have a bit of cleanup to do, but going to put up this PR first - basically the equivalent of the CLI dev change but for deploy. Needs updated PRs in both `airport` and `cli` that I'm going to put up soon

## Test plan
Ran views init and views deploy using CLI (PR incoming) and verified that it worked. 

---
cc @leeweisberger @stepzhou


pkg/build/view.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	// TODO: patch package.json correctly.
+	return applyTemplate(heredoc.Doc(`
+		FROM {{.Base}} as builder
+		WORKDIR /airplane
+
+		COPY package*.json yarn.* /airplane/
+		RUN [ -f package.json ] || { echo "{}" > package.json; }
+		RUN yarn add vite @vitejs/plugin-react react react-dom @airplane/views
```
zhan-airplane, line 76: Going to have to pick specific minimum versions for these probably

pkg/build/view.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// view creates a dockerfile for an view.
+func view(root string, options KindOptions) (string, error) {
+	// Assert that the entrypoint file exists:
+	entrypoint, _ := options["entrypoint"].(string)
+	if entrypoint == "" {
+		return "", errors.New("expected an entrypoint")
+	}
+	/*if err := fsx.AssertExistsAll(filepath.Join(root, entrypoint)); err != nil {
```
rose-huang, line 22: delete this if we're not using it?

pkg/build/views/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
rose-huang, line 1: can we use the same file for dev and deploy for this, the main.tsx file, and the vite config? the dev one is in cli right? also feel free to change this in another PR

pkg/examples/typescript/yarnworkspaces/yarn.lock
```diff
@@ -1,54 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
rose-huang, line 1: was deleting this intentional?

rose-huang (approved):

pkg/build/view.go
rose-huang: also are we using root for anything?

pkg/build/view.go
```diff
@@ -0,0 +1,120 @@
...
+
+import (
+	_ "embed"
+	"strings"
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/pkg/errors"
+)
+
+func DoView(root string, options KindOptions) (string, error) {
```
rose-huang, line 11: what is this used for?

zhan-airplane: Oops was using this to test this function, will delete

pkg/build/views/index.html
zhan-airplane: Yup I'm planning to shift a bunch of code from cli to lib after this merges

PR #180, opened 2022-07-13
Add support for code-only node task definitions

ameyakhare:
Resolves AIR-4316
Resolves AIR-4318
Resolves AIR-4317

## Description
This adds all of the lib changes for code-only node task definitions. Adds the following:

1) code_task_discoverer: discovers ts files of name *.tasks.ts and *.tasks.js
2) parser: runs analysis on task ts files and extracts task configs
3) changes discover api a little to allow for returning multiple task configs from the same file
4) updates node shim to execute code-only tasks correctly

## Test plan
Tested locally and the whole thing works E2E.


pkg/deploy/discover/code_task_discover.go
```diff
@@ -0,0 +1,157 @@
...
+			EnvSlug: c.EnvSlug,
+		})
+		if err != nil {
+			var merr *api.TaskMissingError
+			if !errors.As(err, &merr) {
+				return nil, errors.Wrap(err, "unable to get task")
+			}
+
+			c.Logger.Warning(`Task with slug %s does not exist, skipping deployment.`, def.Slug)
+			return nil, nil
```
francescaleung, line 106: Does this mean that code-only tasks still need to be created outside of this process? I can't just create a typescript file with arbitrary tasks & deploy it?

pkg/deploy/discover/code_task_discover.go
```diff
@@ -0,0 +1,157 @@
...
+func (c *CodeTaskDiscoverer) IsAirplaneTask(ctx context.Context, file string) (string, error) {
+	// Run thing that extracts slug from something in the file.
+	return "", nil
+}
+
+func (c *CodeTaskDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([]TaskConfig, error) {
+	// Only analyze typescript files that end with .tasks.ts
+	if !strings.HasSuffix(file, ".tasks.ts") {
+		return nil, nil
+	}
```
francescaleung, line 36: Is javascript not supported?

pkg/deploy/discover/code_task_discover.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// Run parser on the file
+	out, err := exec.Command("npx", "ts-node", "parser/node/parser.ts", file).Output()
+	if err != nil {
+		return nil, err
+	}
```
francescaleung, line 42: `npx` is a prerequisite for code-only node tasks? where is parser.ts coming from?

pkg/build/node.go
```diff
@@ -144,18 +144,34 @@ func node(root string, options KindOptions, buildArgs []string) (string, error)
 	}
 	cfg.InlineShimPackageJSON = inlineString(string(pjson))
 
+	entrypointFunc, _ := options["entrypointFunc"].(string)
+	paramSlugs, _ := options["paramSlugs"].([]string)
```
francescaleung, line 148: you should have checks on the second return value here

pkg/deploy/discover/code_task_discover.go
```diff
@@ -0,0 +1,157 @@
...
+
+		parameters := parsedTask["parameters"].([]interface{})
+		var params []definitions.ParameterDefinition_0_3
+		for _, taskParam := range parameters {
+			constructedParam := taskParam.(map[string]interface{})
+			params = append(params, definitions.ParameterDefinition_0_3{
+				Name: constructedParam["name"].(string),
+				Slug: constructedParam["name"].(string),
+				Type: constructedParam["type"].(string),
+			})
```
francescaleung, line 121: no support for options/regex/required/default?

francescaleung (commented):
leaving comments here for now, I think I might benefit from a walkthrough &/ pointers to other relevant bits of code that this depends on.

Lots of comments here that aren't intended to be a "we should implement this" but rather a "is this a known limitation". Known limitations are fine as long as they're documented somewhere.

pkg/deploy/discover/code_task_discover.go
ameyakhare: javascript is supported now!

pkg/deploy/discover/code_task_discover.go
ameyakhare: npx is a prerequisite, but nothing else is. parser.ts is coming from an embedded string that is written to a temp file and executed

pkg/deploy/discover/code_task_discover.go
ameyakhare: we have support for all of those now

pkg/build/node-shim.js
```diff
@@ -1,6 +1,10 @@
 // This file includes a shim that will execute your task code.
 import airplane from "airplane";
+{{if .IsCodeDefinedTask}}
+import { {{.EntrypointFunc}} } from "{{.Entrypoint}}";
```
colinking, line 4: You can rename the import here, so you don't need to change the code below. It also avoids any issues with the entrypoint's name clashing with other variable definitions in this file.

```suggestion
import { {{.EntrypointFunc}} as task } from "{{.Entrypoint}}";
```

pkg/build/node-shim.js
```diff
@@ -15,7 +19,11 @@ async function main() {
   }
 
   try {
+    {{if .IsCodeDefinedTask}}
+    let ret = await {{.EntrypointFunc}}._airplane.baseFunc(JSON.parse(process.argv[2]));
```
colinking, line 23:
```suggestion
    let ret = await task._airplane.baseFunc(JSON.parse(process.argv[2]));
```

colinking, line 23: Minor: can we do double-underscores to match the convention of internal fields elsewhere? (e.g. `__airplaneType`).

```suggestion
    let ret = await {{.EntrypointFunc}}.__airplane.baseFunc(JSON.parse(process.argv[2]));
```

pkg/build/node-shim.js
```diff
@@ -15,7 +19,11 @@ async function main() {
   }
 
   try {
+    {{if .IsCodeDefinedTask}}
```
colinking, line 22: You could drop this templating logic and instead check the type of `task` -- `if (task instanceof airplane.Task) {...}`. That should allow us to drop the `isCodeDefinedTask` field from the build options?

pkg/build/node.go
```diff
@@ -144,18 +144,29 @@ func node(root string, options KindOptions, buildArgs []string) (string, error)
 	}
 	cfg.InlineShimPackageJSON = inlineString(string(pjson))
 
+	entrypointFunc, ok := options["entrypointFunc"].(string)
+	if !ok {
+		entrypointFunc = "task"
+	}
```
colinking, line 150: Why do we default this to `task`?

pkg/deploy/discover/code_task_discover.go
```diff
@@ -0,0 +1,173 @@
...
+type CodeTaskDiscoverer struct {
+	Client  api.IAPIClient
+	Logger  logger.Logger
+	EnvSlug string
+}
+
+var _ TaskDiscoverer = &CodeTaskDiscoverer{}
+
+func (c *CodeTaskDiscoverer) IsAirplaneTask(ctx context.Context, file string) (string, error) {
+	// Run thing that extracts slug from something in the file.
```
colinking, line 33: Shouldn't this be checking if the file ends in `.tasks.[js|ts]`?

pkg/deploy/discover/code_task_discover.go
```diff
@@ -0,0 +1,173 @@
...
+	return "", nil
+}
+
+func (c *CodeTaskDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([]TaskConfig, error) {
+	if !(strings.HasSuffix(file, ".tasks.ts") || strings.HasSuffix(file, ".tasks.js")) {
+		return nil, nil
+	}
```
colinking, line 40: Minor: could we do `.task.ts` and `.task.js`? Singular instead of plural. That lines up with the common case of a file having a single task.

pkg/deploy/discover/code_task_discover.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Create a temp file
+	tempFile, err := os.CreateTemp("", "airplane.parser.node.*.ts")
+	if err != nil {
+		return nil, nil
+	}
```
colinking, line 46: Add a defer statement to remove this file.

pkg/deploy/discover/code_task_discover.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Run parser on the file
+	out, err := exec.Command("npx", "-p", "typescript", "-p", "@types/node", "-p", "ts-node",
+		"ts-node", tempFile.Name(), file).Output()
+	if err != nil {
+		return nil, err
+	}
+
+	var parsedTasks []map[string]interface{}
```
colinking, line 59: Let's replace the map with a proper struct definition.

pkg/deploy/discover/code_task_discover.go
```diff
@@ -0,0 +1,173 @@
...
+			Slug:    def.Slug,
+			EnvSlug: c.EnvSlug,
+		})
+		if err != nil {
+			var merr *api.TaskMissingError
+			if !errors.As(err, &merr) {
+				return nil, errors.Wrap(err, "unable to get task")
+			}
+
+			c.Logger.Warning(`Task with slug %s does not exist, skipping deployment.`, def.Slug)
```
colinking, line 120: Let's expose similar logic as the definition deployer to automatically deploy tasks that do not exist: https://github.com/airplanedev/lib/blob/25e800a19d255ef509e6ac9f2e06f3fa886c9e78/pkg/deploy/discover/defn_discover.go#L78-L98

pkg/deploy/discover/code_task_discover.go
```diff
@@ -0,0 +1,173 @@
...
+		def.SetBuildConfig("entrypoint", ep)
+		def.SetBuildConfig("entrypointFunc", parsedTask["entrypointFunc"].(string))
+		def.SetBuildConfig("isCodeDefinedTask", true)
+
+		var paramSlugs []string
+		for _, param := range def.Parameters {
+			// Preserve order of parameters
+			paramSlugs = append(paramSlugs, param.Slug)
+		}
+		def.SetBuildConfig("paramSlugs", paramSlugs)
```
colinking, line 153: Is this new? What is it for?

pkg/deploy/discover/code_task_discover.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	absEntrypoint, err := pathcase.ActualFilename(absFile)
+	if err != nil {
+		return nil, err
+	}
+	ep, err := filepath.Rel(taskroot, absEntrypoint)
+	if err != nil {
+		return nil, err
+	}
```
colinking, line 92: IIRC, we have similar logic elsewhere. Can we extract + share this?

pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -0,0 +1,82 @@
...
+type ParamSchema = {
+    name: string
+    kind: string
+}
+
+type TaskConfig = {
+    slug: string
+    name: string
+    description?: string
+    parameters?: Record<string, ParamSchema>
```
colinking, line 12: Go does not retain key ordering when deserializing maps (but JS does). You might need to use [`ojson`](https://github.com/airplanedev/ojson)? Or serialize this as a list.

pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -0,0 +1,82 @@
...
+    timeout?: number
+    constraints?: Record<string, string>
+    runtime?: "standard" | "workflow"
+}
+
+type TaskConfigWithBuildArgs = TaskConfig & {
+    entrypointFunc: string
+}
+
+class AirplaneParser {
```
colinking, line 24: Seems simpler to express this as a function rather than a class:

```ts
const extractTaskConfigs = (files: string[]): TaskConfig[] => {
  // ...
}
```

pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    extractTaskConfigs(): TaskConfig[] {
+        // Get function signatures and their parameter types
+        // this.extractFSignatures();
+
+        // Import each of the files
+        let configs: TaskConfigWithBuildArgs[] = [];
+        for (const file of this.files) {
+            const resolvedPath = path.relative(__dirname, file);
+            const lib = resolvedPath.replace(RegExp(".ts$"), "");
```
colinking, line 39:
```suggestion
            const lib = resolvedPath.replace(/.ts$/, "");
```

pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -0,0 +1,82 @@
...
+            let imports = require(`./${lib}`);
```
colinking, line 40:
```suggestion
            const exports = require(`./${lib}`);
```

pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -0,0 +1,82 @@
...
+
+            for (const itemName in imports) {
```
colinking, line 42:
```suggestion
            for (const exportName in exports) {
```

pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -0,0 +1,82 @@
...
+                const item = imports[itemName];
```
colinking, line 43:
```suggestion
                const export = exports[exportName];
```

pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -0,0 +1,82 @@
...
+                
+                if ("_airplane" in item) {
+                    const config = item._airplane.config;
+
+                    var params: Record<string, ParamSchema> = {};
+                    for (var uParamSlug in config.parameters) {
+                        const uParamConfig = config.parameters[uParamSlug];
+
+                        if (typeof uParamSlug === "string") {
```
colinking, line 52: When would the slug not be a string?

colinking (changes requested):

pkg/deploy/discover/code_task_discover.go
ameyakhare: will be done in a follow up (another task assigned to me) but as of now, yes

pkg/deploy/discover/parser/node/parser.ts
ameyakhare: oops meant to check uParamConfig not uParamSlug

pkg/build/node.go
ameyakhare: removed default, should work with default export and non-default export now

pkg/build/node-shim.js
ameyakhare: I just checked for .__airplane on the task until I land the node-sdk changes

pkg/build/node.go
ameyakhare: no longer required

pkg/build/node-shim.js
colinking: Ah, nm about this. This only works if `airplane.Task` is a class. What you have now is fine!

pkg/deploy/discover/code_task_discover.go
ameyakhare: we don't need this anymore

pkg/deploy/discover/parser/node/parser.ts
ameyakhare: yeah this is an artifact from when parsing was a little more complicated

pkg/deploy/discover/code_task_discover.go
colinking: FWIW, npx is installed automatically with npm/node nowadays. If you have node installed, you'll have npx, too.

pkg/deploy/discover/parser/node/parser.ts
ameyakhare: serialized as list to match definitions reference

pkg/deploy/discover/code_task_discover.go
ameyakhare: im outputting json in the definitions reference format so we dont actually need to do any parsing ourselves 🎉

pkg/deploy/discover/code_task_discover.go
```diff
@@ -66,99 +76,84 @@ func (c *CodeTaskDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([
...
-		if task.IsArchived {
-			c.Logger.Warning(`Task with slug %s is archived, skipping deployment.`, def.Slug)
-			return nil, nil
+
+		if err := def.Unmarshal(definitions.DefFormatJSON, b); err != nil {
+			switch err := errors.Cause(err).(type) {
+			case definitions.ErrSchemaValidation:
+				errorMsgs := []string{}
+				for _, verr := range err.Errors {
+					errorMsgs = append(errorMsgs, fmt.Sprintf("%s: %s", verr.Field(), verr.Description()))
```
colinking, line 106: 🔥

pkg/deploy/discover/code_task_discover.go
```diff
@@ -66,99 +76,84 @@ func (c *CodeTaskDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([
...
 
-		var paramSlugs []string
-		for _, param := range def.Parameters {
-			// Preserve order of parameters
-			paramSlugs = append(paramSlugs, param.Slug)
-		}
-		def.SetBuildConfig("paramSlugs", paramSlugs)
+			if c.MissingTaskHandler == nil {
+				return nil, nil
+			}
```
colinking, line 135: Let's skip just the missing task, but continue with the other tasks that exist!

pkg/deploy/discover/code_task_discover.go
```diff
@@ -66,99 +76,84 @@ func (c *CodeTaskDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([
...
-		if err := def.SetWorkdir(taskroot, wd); err != nil {
-			return nil, err
+			mptr, err := c.MissingTaskHandler(ctx, &def)
+			if err != nil {
+				return nil, err
+			} else if mptr == nil {
+				if c.Logger != nil {
+					c.Logger.Warning(`Task with slug %s does not exist, skipping deployment.`, def.GetSlug())
+				}
+				return nil, nil
```
colinking, line 144: Same here, let's continue to process the other tasks.

pkg/deploy/discover/code_task_discover.go
```diff
@@ -66,99 +76,84 @@ func (c *CodeTaskDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([
...
+			}
+			metadata = *mptr
+		}
+		if metadata.IsArchived {
+			c.Logger.Warning(`Task with slug %s is archived, skipping deployment.`, metadata.Slug)
+			return nil, nil
```
colinking, line 150: Here, too.

pkg/deploy/discover/code_task_discover.go
```diff
@@ -66,99 +76,84 @@ func (c *CodeTaskDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([
...
 
-		if err := def.SetAbsoluteEntrypoint(absFile); err != nil {
+		if err := def.SetAbsoluteEntrypoint(pathMetadata.AbsFile); err != nil {
 			return nil, err
 		}
 
-		def.SetBuildConfig("entrypoint", ep)
-		def.SetBuildConfig("entrypointFunc", parsedTask["entrypointFunc"].(string))
+		// Task metadata
+		metadata, err := c.Client.GetTaskMetadata(ctx, def.GetSlug())
```
colinking, line 126: nice

pkg/deploy/discover/script_discover.go
```diff
@@ -59,51 +60,72 @@ func (sd *ScriptDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([]
...
+
+func (sd *ScriptDiscoverer) ConfigSource() ConfigSource {
+	return ConfigSourceScript
+}
+
+type TaskPathMetadata struct {
+	AbsFile    string
+	Entrypoint string
+	TaskRoot   string
+	Workdir    string
```
colinking, line 91: Minor, these names are more clear IMO:

```suggestion
	AbsEntrypoint    string
	RelEntrypoint string
	RootDir   string
	WorkDir    string
```

pkg/deploy/discover/script_discover.go
```diff
@@ -59,51 +60,72 @@ func (sd *ScriptDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([]
...
+}
+
+func taskPathMetadata(file string, kind build.TaskKind) (TaskPathMetadata, error) {
```
colinking, line 94: Extract from `discoverer/script_discoverer.go` to... `discoverer/path.ts`?

pkg/deploy/discover/script_discover.go
```diff
@@ -59,51 +60,72 @@ func (sd *ScriptDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([]
 		return nil, err
 	}
 
-	r, err := runtime.Lookup(file, task.Kind)
+	pathMetadata, err := taskPathMetadata(file, task.Kind)
```
colinking, line 63: Nice. Add to definition discoverer too: https://github.com/airplanedev/lib/blob/03140289bf0837d36d4ddf83420e969726118b88/pkg/deploy/discover/defn_discover.go#L121-L138

colinking (changes requested):
LGTM apart from a question is about the `IsAirplaneTask` API

pkg/deploy/discover/code_task_discover.go
```diff
@@ -4,38 +4,47 @@ import (
...
-	// Run thing that extracts slug from something in the file.
-	return "", nil
+	taskConfigs, err := c.GetTaskConfigs(ctx, file)
+	if err != nil {
+		return "", err
+	}
+	if len(taskConfigs) == 0 {
+		return "", nil
+	}
+	return taskConfigs[0].Def.GetSlug(), nil
```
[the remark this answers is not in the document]
ameyakhare: changed it to GetAirplaneTasks cause its returning a list and not a bool

pkg/deploy/discover/code_task_discover.go
```diff
@@ -66,99 +76,84 @@ func (c *CodeTaskDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([
...
 	var taskConfigs []TaskConfig
 	for _, parsedTask := range parsedTasks {
-		// Construct task config
-		def := definitions.Definition_0_3{
-			Name: parsedTask["name"].(string),
-			Slug: parsedTask["slug"].(string),
-			Node: &definitions.NodeDefinition_0_3{
-				NodeVersion: "18",
-			},
+		// Modify fields to add ones that are strictly required
```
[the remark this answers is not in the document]
ameyakhare: I need to add the node: {entrypoint: foo, nodeVersion: foo} stanza anyways, so gonna skip this.

pkg/deploy/discover/code_task_discover.go
```diff
@@ -66,99 +76,84 @@ func (c *CodeTaskDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([
...
+		entrypointFunc := parsedTask["entrypointFunc"].(string)
+		delete(parsedTask, "entrypointFunc")
+		parsedTask["node"] = map[string]interface{}{
+			"nodeVersion": "18",
```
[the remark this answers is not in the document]
ameyakhare: I don't think you can leave it unset, the schema_0_3 requires it to be set. made a const on node.go in the build package

colinking (approved):

PR #185, opened 2022-07-18
Create builtins package

kellymtrinh:
## Description
Creates a new package for builtins related code that's needed across multiple repos (internal/airport and cli, for example).
For now, we only need the slug logic and resource structs.

## Test plan
This code is unused for now.


pkg/builtins/builtins.go
```diff
@@ -0,0 +1,37 @@
+package builtins
+
+import (
+	"fmt"
+	"strings"
+)
+
+const builtinsSlugPrefix = "airplane"
+
```
kellymtrinh, line 9: This is all from https://github.com/airplanedev/airport/blob/4eb81c9e0fb7a55d9d1ba82918684706ee6bead8/pkg/runs/builtins.go#L17-L40 which will need a followup PR to use the lib package instead

francescaleung (approved):

PR #186, opened 2022-07-19
Allow resources to be a list

francescaleung:
## Description

Resource attachments can now be specified as a list, where each element is either a resource slug (where alias = slug) or an object with a single key (where key = alias, value = slug).

```
resources:
  - httpbin
  - db: demo_db
```

When running `airplane init`, resource attachments are preferred as a mapping of alias -> slug.

```
resources:
  httpbin: httpbin
  db: demo_db
```

## Test plan

Unit tests, tested manually.

### DIFF
--- a/pkg/deploy/taskdir/definitions/schema_0_3.json
+++ b/pkg/deploy/taskdir/definitions/schema_0_3.json
@@ -404,7 +404,7 @@
             {
               "type": "array",
               "items": {
-                { "type": "string", "pattern": "^[a-z0-9_]{1,50}$" }
+                "type": "string", "pattern": "^[a-z0-9_]{1,50}$"
               }
             }
           ]
--- a/pkg/deploy/taskdir/definitions/schema_0_3.json
+++ b/pkg/deploy/taskdir/definitions/schema_0_3.json
@@ -390,7 +390,7 @@
           }
         },
         "resources": {
-          "description": "[Experimental] A map of resources that can be accessed from your task. Keys are aliases, which determine how you refer to the resource in your task; values are existing resource slugs in your Airplane team. Can also be a list of resource slugs (where the alias is the slug) or a list of objects with a single alias -> slug mapping.",
+          "description": "[Experimental] A list of resources to make available to your task. Resources are identified by slug, but can be mapped to an alias to configure how the task references the resource. If using aliases, resources are expressed as a map of alias to slug.",
           "oneOf": [
             {
               "type": "object",
--- a/pkg/deploy/taskdir/definitions/schema_0_3.json
+++ b/pkg/deploy/taskdir/definitions/schema_0_3.json
@@ -390,7 +390,7 @@
           }
         },
         "resources": {
-          "description": "[Experimental] A list of resources to make available to your task. Resources are identified by slug, but can be mapped to an alias to configure how the task references the resource. If using aliases, resources are expressed as a map of alias to slug.",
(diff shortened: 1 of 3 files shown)


pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -390,14 +390,37 @@
           }
         },
         "resources": {
-          "description": "[Experimental] A map of resources that can be accessed from your task. Keys determine how you refer to the resource; values are existing resource slugs in your Airplane team.",
-          "type": "object",
-          "patternProperties": {
-            "^[a-z0-9_]{1,50}$": {
-              "type": "string",
-              "pattern": "^[a-z0-9_]{1,50}$"
+          "description": "[Experimental] A map of resources that can be accessed from your task. Keys are aliases, which determine how you refer to the resource in your task; values are existing resource slugs in your Airplane team. Can also be a list of resource slugs (where the alias is the slug) or a list of objects with a single alias -> slug mapping.",
```
colinking, line 393: Can we remove the `[Experimental]` tag?

colinking, line 393: IMO it's more clear to explain w/out aliases first:

```suggestion
          "description": "A list of resources to make available to your task. Resources are identified by slug, but can be mapped to an alias to configure how the task references the resource. If using aliases, resources are expressed as a map of alias to slug.",
```

colinking (changes requested):
Looks good! Let's make the list of slugs the default serialization method when initing a task, so that aliases are not exposed unless a developer is using aliases.

Also, the list of objects is clever! Took me a second to grok what was happening. TBH, I'd prefer there weren't two identical ways to express aliased resources:

```yaml
resources:
  db: airplane_db
  api: airplane_api
```

```yaml
resources:
  - db: airplane_db
  - api: airplane_api
```

Can we remove this overlap by either dropping map support or dropping list-of-objects support?

francescaleung:
> Looks good! Let's make the list of slugs the default serialization method when initing a task, so that aliases are not exposed unless a developer is using aliases.
> 
> Also, the list of objects is clever! Took me a second to grok what was happening. TBH, I'd prefer there weren't two identical ways to express aliased resources:
> 
> ```yaml
> resources:
>   db: airplane_db
>   api: airplane_api
> ```
> 
> ```yaml
> resources:
>   - db: airplane_db
>   - api: airplane_api
> ```
> 
> Can we remove this overlap by either dropping map support or dropping list-of-objects support?

Pretty sure we talked about list-of-objects a month or so ago, ha. Does it matter that we support both? I would prefer to drop list-of-objects because we already support map & I don't want to remove that.

pkg/deploy/taskdir/definitions/schema_0_3.json
francescaleung: `[Experimental]` is because resource attachment never got officially released--the flag is still on in prod that prevents people from seeing the resource attachment UI. It's tagged as `[Experimental]` because this schema file is the one that feeds the intellisense in y'all's fancy code editors as long as it's in main.

It never got officially released because there's not much of a point in attaching resources if you can't actually use them (documentation, but mostly was waiting on SDK).

colinking:
> Pretty sure we talked about list-of-objects a month or so ago, ha. Does it matter that we support both? I would prefer to drop list-of-objects because we already support map & I don't want to remove that.

Dropping lists-of-objects sgtm!

colinking (changes requested):
Requesting changes about the experimental tag + dropping list-of-objects

colinking (approved):

PR #187, opened 2022-07-19
Ensure name is valid YAML

francescaleung:
## Description

This only really needs to be done for name, as it's one of the few freeform fields that folks get when initializing a task. (Entrypoint is the other one, but not sticking to ascii characters for entrypoints seems like asking for trouble. Willing to punt on this.)

Resolves AIR-4252.

## Test plan

Unit tests.

### DIFF
--- a/pkg/deploy/taskdir/definitions/def_0_3.go
+++ b/pkg/deploy/taskdir/definitions/def_0_3.go
@@ -837,6 +837,15 @@ func (d Definition_0_3) GenerateCommentedFile(format DefFormat) ([]byte, error) 
 		return d.Marshal(format)
 	}
 
+	// Run yaml.Marshal to take care of any weird characters. Note that this won't work if there
+	// are newlines in the name.
+	nameBuf, err := yaml.Marshal(d.Name)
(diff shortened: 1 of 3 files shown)


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -837,14 +837,23 @@ func (d Definition_0_3) GenerateCommentedFile(format DefFormat) ([]byte, error)
 		return d.Marshal(format)
 	}
 
+	// Run yaml.Marshal to take care of any weird characters. Note that this won't work if there
+	// are newlines in the name.
+	nameBuf, err := yaml.Marshal(d.Name)
```
ameyakhare, line 842: can we get rid of newlines with strings.ReplaceAll?

ameyakhare (approved):

pkg/deploy/taskdir/definitions/def_0_3.go
francescaleung: ooh fair, especially since newlines in a name are weird anyway.

PR #188, opened 2022-07-19
Clear yarn cache after installing

calufornia:
## Description

I've noticed that when deploying some workflow tasks, the image size is very large (~1.9GB), which causes pushing to artifact registry to take a significant amount of time. Most of the size seems to come from the yarn cache:
```
root@5a2283c9cd08:/usr/local/share/.cache# du -sh yarn
1.5G	yarn
```
Unfortunately, yarn doesn't seem to have an option to not use a cache in the first place, and so the best we can do is remove it in the same RUN instruction that we install the packages in.

## Test plan

Tested by deploying a task with the cli pointing to a local version of lib, and verifying that the size of the resulting image was reduced (~1.9GB -> 450MB).

### DIFF
--- a/pkg/build/node.go
+++ b/pkg/build/node.go
@@ -177,7 +177,6 @@ func node(root string, options KindOptions, buildArgs []string) (string, error) 
 			// cache directory for storing packages.
 			installCommand = "yarn install --non-interactive --production --frozen-lockfile && yarn cache clean"
 		}
-
 	} else if hasPackageLock {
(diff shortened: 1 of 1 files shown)


calufornia:
@leeweisberger The yarn cache isn't used for anything after we `yarn install ...`, right? I'm thinking that this should be safe to do?

leeweisberger (approved):
lgtm

PR #193, opened 2022-07-21
Respect existing package.json packages on build

zhan-airplane:
## Description
Didn't realize `yarn add` would put in the latest version even if it already exists in the `package.json`, so this handles that case

## Test plan
Ran locally

---
cc @stepzhou @rose-huang


pkg/build/view.go
```diff
@@ -67,14 +67,18 @@ func view(root string, options KindOptions) (string, error) {
...
+		RUN apt update && apt install -y jq
 
 		COPY package*.json yarn.* /airplane/
 		RUN [ -f package.json ] || { echo "{}" > package.json; }
-		RUN yarn add vite @vitejs/plugin-react react react-dom @airplane/views
+		RUN [ $(jq .dependencies.vite package.json) != 'null' ] || yarn add vite
+		RUN [ $(jq .dependencies.\"@vitejs/plugin-react\" package.json) != 'null' ] || yarn add @vitejs/plugin-react
+		RUN [ $(jq .dependencies.react package.json) != 'null' ] || yarn add react
+		RUN [ $(jq .dependencies.react-dom package.json) != 'null' ] || yarn add react-dom
+		RUN [ $(jq .dependencies.\"@airplane/views\" package.json) != 'null' ] || yarn add @airplane/views
```
leeweisberger, line 81: This is going to run these `yarn add` in serial right? We want to run them in parallel.

Rather than using jq - how about doing this in Go code above. Grab the package.json and convert it to a struct (we do this in lib in a few places). Decide which packages we need to install. Insert them into this template using a go template string `RUN yarn add {{.packagesToAdd}}` (although we can skip if packagestoadd is empty

leeweisberger (changes requested):

pkg/build/view.go
zhan-airplane: Ah yeah that's a good idea I'll do that instead, somehow forgot I could write it in go oops haha

pkg/build/view.go
```diff
@@ -49,6 +51,29 @@ func view(root string, options KindOptions) (string, error) {
...
+			return "", errors.Wrap(err, "reading package JSON")
+		}
+		if err := json.Unmarshal([]byte(packageJSONFile), &packageJSON); err != nil {
+			return "", errors.Wrap(err, "parsing package JSON")
+		}
+	}
+
+	packagesToCheck := []string{"vite", "@vitejs/plugin-react", "react", "react-dom", "@airplane/views"}
+	packagesToAdd := []string{}
+	deps, ok := packageJSON.(map[string]interface{})["dependencies"].(map[string]interface{})
```
leeweisberger, line 68: Check dev dependencies too. the vite stuff might very well be in devdependencies

pkg/build/view.go
```diff
@@ -65,16 +91,18 @@ func view(root string, options KindOptions) (string, error) {
...
 	return applyTemplate(heredoc.Doc(`
 		FROM {{.Base}} as builder
 		WORKDIR /airplane
 
 		COPY package*.json yarn.* /airplane/
 		RUN [ -f package.json ] || { echo "{}" > package.json; }
-		RUN yarn add vite @vitejs/plugin-react react react-dom @airplane/views
+		{{if .PackagesToAdd }}
+			RUN yarn add {{.PackagesToAdd}}
+		{{end}}
```
leeweisberger, line 105: Rather than doing an add and then an install, it would be more performant to add the dependencies to the package.json and then run the install command. This would change from 2 installs to 1. Would pretty pretty easy - just add to the dependency list and then you can remove this line. I'll leave it up to you if you want to do this now

pkg/build/view.go
```diff
@@ -57,6 +82,7 @@ func view(root string, options KindOptions) (string, error) {
 		InlineIndexHtml  string
 		InlineViteConfig string
 		APIHost          string
+		PackagesToAdd    string
```
leeweisberger, line 85: While you're here, can you update the InstallCommand to clean the cache. a'la https://github.com/airplanedev/lib/pull/188/files

leeweisberger (approved):

zhan-airplane:
@leeweisberger actually while messing with this I have a question about `--frozen-lockfile` - if we're manually editing the `package.json`, `--frozen-lockfile`'s always going to fail right..?

zhan-airplane:
Is it ok to get rid of it? Not too sure what the best practice would be here

PR #199, opened 2022-07-27
Add custom json marshaller for resource attachments

calufornia:
## Description

Something I noticed when testing https://github.com/airplanedev/cli/pull/540 - we currently have a custom JSON unmarshaler for resource attachments, but not a marshaler. This causes resources attachments to by default get (JSON) marshaled as
```
...
"resources": {
  "attachments": {
    "db": "demo_db"
  }
}
...
```
This subsequently causes unmarshaling to fail since a map of maps is unexpected. In actuality, resource attachments should be marshaled as 
```
...
"resources": {
  "db": "demo_db"
}
...
```
This isn't a problem today because there _is_ a `MarshalYAML` method that properly marshals the definition into yaml for deploys (we use it for json too) - this is mainly a problem in unit tests that don't use a yaml marshaler.

## Test plan
Tested by deploying a task with resource attachments locally - also unit tests in above PR.


francescaleung:
ooh good catch. Can you try removing the MarshalYAML function & see if it still works as intended for YAML?

I tried so hard to only have one set of functions & still messed it up, oh no.

calufornia:
> ooh good catch. Can you try removing the MarshalYAML function & see if it still works as intended for YAML?
> 
> I tried so hard to only have one set of functions & still messed it up, oh no.



> ooh good catch. Can you try removing the MarshalYAML function & see if it still works as intended for YAML?
> 
> I tried so hard to only have one set of functions & still messed it up, oh no.

So interestingly enough, it seems like we still need the `MarshalYAML` in order for the go-yaml package to properly marshal the definition into yaml/json. If we remove `MarshalYaml`, the yaml file looks like:
```
resources:
demo: db
```
which makes it seem like the map entries are getting marshaled as top-level fields. Additionally, even if we do something like
```
return json.Marshal(struct {
	Resources map[string]string `json:"resources"`
}{
	Resources: r.Attachments,
})
```
the yaml output is:
```
resources:
resources:
    demo: db
```
This seems like a bug with the go-yaml package around marshaling map types using a custom MarshalJSON method.

francescaleung:
bummer. Can you leave a comment indicating why we need both MarshalJSON + MarshalYAML?

francescaleung (approved):

PR #202, opened 2022-07-28
Change error output format for Node tasks

colinking:
## Description

If a Node task threw an error, it would get outputted:

```ts
throw new Error("oops!")
// ->
{
	"error": [
		{"error": "Error: oops!"}
	]
}
```

After this change, errors will (1) have one less level of wrapping and (2) the top-level is switched from a list to a singular error:

```ts
throw new Error("oops!")
// ->
{
	"error": "Error: oops!"
}
```

This is a breaking change as the output format will change for Node tasks.  Since users cannot access output of failed tasks from runbooks, this does not impact runbook usage. It only impacts programmatic usage of tasks. It additionally impacts users who are throwing errors _and_ producing output to the `.error` key.

This change also aligns Node task errors with built-in function errors, e.g. an example from a SQL task:

```json
{
  "error": "ErrBadRequest: dial tcp 35.223.27.46:5432: connect: connection timed out"
}
```

After this merges, we'll update our Node task documentation to clarify the error handling semantics.

cc @ericliuche 

## Test plan

Run a task with a local version of the SDK locally. Throw an error and confirm it is produced correctly.


colinking:
* main
  * **PR #202** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/202" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/202?utm_source=stack-comment).

leeweisberger (approved):
sgtm. I'm in favor

PR #203, opened 2022-07-28
Update workflow shim to produce cancelled/failed statuses

colinking:
## Description

Updates the Node workflow shim to produce `airplane_status:active|cancelled|failed` logs.

Depends on: https://github.com/airplanedev/airport/pull/2899

cc @kellymtrinh 

## Test plan

Run a workflow locally and confirmed cancellation and failure are both handled correctly.

PR #204, opened 2022-07-29
Fix workflow output

colinking:
## Description

Fixes an issue that prevented workflows from producing output from return values.

## Test plan

Run a workflow locally that returns a result as output. The output should be rendered in the UI.


colinking:
* main
  * **PR #204** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/204" target="_blank">Graphite</a>  👈
    * **PR #203** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/203" target="_blank">Graphite</a> 
      * **PR #202** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/202" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/204?utm_source=stack-comment).

pkg/build/workflow/workflow-shim.js
```diff
@@ -17,16 +17,14 @@ export async function __airplaneEntrypoint(params, workflowArgs) {
     }
 
     var result = await task(JSON.parse(params[0]));
```
colinking, line 19: This is the bug ^

`result` was scoped to the `try` so `undefined` was always returned as the workflow's output.

pkg/build/workflow/workflow-shim.js
```diff
@@ -17,16 +17,14 @@ export async function __airplaneEntrypoint(params, workflowArgs) {
...
     logger.info('airplane_output_append:error ' + JSON.stringify({ error: String(err) }));
     logger.info('airplane_output_append:error {"error":"Error executing workflow"}')
-    return
-  }
-
-  if (result !== undefined) {
-    airplane.setOutput(result);
   }
-  logger.info('airplane_status:succeeded');
-  return result;
```
colinking, line 31: There's no need to return the result AFAIK, so I removed it.

pkg/build/workflow/worker-shim.js
```diff
@@ -113,26 +112,17 @@ function activityLog(info, message) {
...
+    console.error(`Expected to receive a single argument (via {{ "{{JSON}}" }}). Task CLI arguments may be misconfigured.`);
     process.exit(1);
   }
 
   try {
-    let ret = await runWorker(JSON.parse(process.argv[2]));
-    if (ret !== undefined) {
-      airplane.setOutput(ret);
-    }
+    await runWorker(JSON.parse(process.argv[2]));
```
colinking, line 123: Pretty sure this is a relic of an old version of the worker, but the worker itself does not produce output.

calufornia (approved):
lgtm - thanks for cleaning up the logs too!

PR #205, opened 2022-07-29
Emit resources and schedules from node code config tasks

ameyakhare:
## Description
We want to allow users to define resources and schedules, this pipes those attributes through to the parsed definition.

## Test plan
Deploys correctly.


pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -20,6 +20,8 @@ type Def = {
     allowSelfApprovals?: boolean
     timeout?: number
     constraints?: Record<string, string>
+    resources: Record<string, string>
```
ameyakhare, line 23: im starting to think we should just import the airplane Task config object here... I think there was a reason I was trying to avoid an external dependency though

colinking (approved):

PR #212, opened 2022-08-04
Update views container spacing props

trijeetm:
## Description
Currently there is a lot of extra space around the view container. This fixes it by matching it to how our web app spaces tasks / runbook in a page.

Linked to this PR in `Views`
https://github.com/airplanedev/views/pull/186


zhan-airplane (approved):
Screenshot from 2022-08-04 16-57-59

Just gonna preemptively approve, but lmk if this is what you're expecting

PR #213, opened 2022-08-04
Support deploying views with code configuration

ameyakhare:
Resolves AIR-4417
Resolves AIR-4416

## Description
Adds support for deploying views with code configuration. Users need to export default airplane.view(config, component) instead of default exporting the component itself.

## Test plan
Deploys properly to stage.


leeweisberger (changes requested):
Will look when we have a test suite. We should be able to write some fixtures and ensure that we are discovering the task correctly pretty easy.

pkg/deploy/discover/build.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	if len(res.OutputFiles) == 0 {
+		return errors.New("esbuild failed to produce output files")
+	}
+	return nil
+}
+
+// Patches node_modules/@airplane/views/package.json in the root directory by adding
+// "main": "index.dummy.js" to allow the library to resolve properly when importing
```
leeweisberger, line 67: Can you add some more description for those that don't know what index.dummy.js is

pkg/deploy/discover/code_task_discover.go
```diff
@@ -102,68 +103,69 @@ type ParsedDefinition struct {
...
-	if err != nil {
+
+	if err := esbuildUserFiles(pm.RootDir); err != nil {
 		return nil, err
 	}
+	defer func() {
+		if err := os.RemoveAll(path.Join(pm.RootDir, ".airplane")); err != nil {
+			c.Logger.Warning("unable to remove temporary directory: %s")
+		}
+	}()
```
leeweisberger, line 129: Where is the .airplane dir being created?

pkg/deploy/discover/code_view_discover.go
```diff
@@ -0,0 +1,134 @@
...
+	if !isCodeViewFile(file) {
+		return nil, nil
+	}
+
+	pm, err := taskPathMetadata(file, build.TaskKindNode)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := esbuildUserFiles(pm.RootDir); err != nil {
```
leeweisberger, line 37: How fast or slow is this? Wondering because we are doing it for every file and we are doing it twice, once for tasks and once for views right?

pkg/deploy/discover/code_view_discover.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if len(parsedConfigs.ViewConfigs) == 0 {
+		return nil, nil
+	}
+	if len(parsedConfigs.ViewConfigs) > 1 {
+		return nil, errors.New(fmt.Sprintf("unable to parse multiple views from %s, do not export more than one view", file))
+	}
+
+	d := definitions.ViewDefinition{}
+	parsedConfigs.ViewConfigs[0]["entrypoint"] = pm.AbsEntrypoint
```
leeweisberger, line 70: Shouldn't each of these be typed? Is it an untyped map?

pkg/deploy/discover/code_view_discover.go
```diff
@@ -0,0 +1,134 @@
...
+		Root:   pm.RootDir,
+	}, nil
+}
+
+func (dd *CodeViewDiscoverer) ConfigSource() ConfigSource {
+	return ConfigSourceCode
+}
+
+func isCodeViewFile(file string) bool {
+	for _, suffix := range []string{".view.tsx", ".view.jsx", ".view.ts", ".view.js"} {
```
leeweisberger, line 128: Why are you checking for .ts and .js files?

leeweisberger (commented):
Leaving a few comments - going to finish reviewing in a few....

pkg/deploy/discover/discover.go
```diff
@@ -103,6 +104,9 @@ func (d *Discoverer) Discover(ctx context.Context, paths ...string) ([]TaskConfi
 		}
 
 		if fileInfo.IsDir() {
+			if fileInfo.Name() == ".airplane" {
+				continue
+			}
```
leeweisberger, line 109: Why do we need this when you added .airplane to ignoredDirectories

pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -41,49 +53,59 @@ const extractTaskConfigs = (files: string[]): DefWithBuildArgs[] => {
 
       if ("__airplane" in item) {
         const config = item.__airplane.config;
+        if (item.__airplane.type === "view") {
+          viewConfigs = viewConfigs.concat({
+            slug: config.slug,
+            name: config.name,
+            description: config.description,
+          })
```
leeweisberger, line 61: nit: viewConfigs.push

pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -41,49 +53,59 @@ const extractTaskConfigs = (files: string[]): DefWithBuildArgs[] => {
...
-          allowSelfApprovals: config.allowSelfApprovals,
-          timeout: config.timeout,
-          constraints: config.constraints,
-          runtime: config.runtime === "workflow" ? "workflow" : "",
-          resources: config.resources,
-          schedules: config.schedules,
-          parameters: params,
-          entrypointFunc: exportName,
-        });
+          taskConfigs = taskConfigs.concat({
```
leeweisberger, line 87: push?

pkg/deploy/discover/fixtures/view/codeOnly.view.tsx
```diff
@@ -0,0 +1,10 @@
+export default {
+    __airplane: {
+        type: "view",
+        config: {
+            slug: "my_view",
+            name: "My View",
+            description: "Test view yaml file"
+        }
+    }
+}
```
leeweisberger, line 10: Is this how someone specifies a view? What about their source code? They export that but not as default?

leeweisberger (changes requested):
Just have some qs. Will approve when resolved. Nothing major

pkg/deploy/discover/fixtures/view/codeOnly.view.tsx
leeweisberger: How do we know what the entrypoint component is?

pkg/deploy/discover/code_task_discover.go
ameyakhare: in the root directory of the project

pkg/deploy/discover/code_view_discover.go
ameyakhare: decently slow but can be faster if all files parsed at once

pkg/deploy/discover/fixtures/view/codeOnly.view.tsx
ameyakhare: this is how you define it, but I emulated what the lib did so that we don't have to install it. export default airplane.view({slug:"foo"}, Component);

leeweisberger (approved):

pkg/deploy/discover/discover.go
ameyakhare: fixed

pkg/deploy/discover/code_view_discover.go
ameyakhare: we need to add a field (entrypoint) then validate.

PR #216, opened 2022-08-08
Proxy console.* functions in workflows

calufornia:
## Description
The `console` object is not automatically included in the workflow runtime due to determinism concerns - similar to `process.env`, we patch the `console.*` functions using [logging sinks](https://docs.temporal.io/typescript/logging/) so that a user can utilize these functions in their workflow code. Essentially, each console.* function is mapped to its appropriate entry in the logger sink, which will call the actual console.* function from the worker thread (which does not run in a node isolate). 

## Test plan
Tested locally with the following task:
```
export default async function(params) {
  console.log("log");
  console.info("info");
  console.debug("debug");
  console.warn("warn");
  console.error("error");
}
```


pkg/build/workflow/worker-shim.js
```diff
@@ -46,24 +46,60 @@ async function runWorker(params) {
 
   // Sinks allow us to log from workflows.
   const sinks = {
+    // We prefix all logs with the Temporal workflow ID (equivalent to the Airplane run ID) and the Temporal run ID so
+    // we can link the logs back to specific Airplane and Temporal runs.
     logger: {
+      debug: {
+        fn(workflowInfo, message) {
+          console.debug(
+              `airplane_workflow_log:workflow//${workflowInfo.workflowId}/${workflowInfo.runId} ${message}`
```
colinking, line 55: Extract the `airplane_workflow_log` generation into a shared function. It helps convey that all of these functions produce the same log format.

pkg/build/workflow/worker-shim.js
```diff
@@ -46,24 +46,60 @@ async function runWorker(params) {
 
   // Sinks allow us to log from workflows.
   const sinks = {
+    // We prefix all logs with the Temporal workflow ID (equivalent to the Airplane run ID) and the Temporal run ID so
+    // we can link the logs back to specific Airplane and Temporal runs.
     logger: {
+      debug: {
+        fn(workflowInfo, message) {
+          console.debug(
```
colinking, line 54: Is there any distinction between using `console.debug|warn|...` vs. always using `console.log`? Or is the distinction not relevant to us?

colinking (approved):

pkg/build/workflow/worker-shim.js
calufornia: Yeah I don't think there's too much of a distinction, since these logs aren't being used in the developer tab (and hence don't get special highlighting) - this would actually be a nice way to enhance our own logs.

PR #219, opened 2022-08-09
Register workflow runtime in shim

calufornia:
## Description

As a follow-up to https://github.com/airplanedev/node-sdk/pull/87, registers the workflow runtime in the node shim since this is now separated from the `airplane` package.

## Test plan
Tested with a workflow task locally that did not install `@airplane/workflow-runtime`, and verified that the workflow runtime was registered and the workflow task executed successfully.


pkg/build/workflow/workflow-shim.js
```diff
@@ -9,6 +10,7 @@ const { logger } = proxySinks();
 // This name must match the name we use when executing the workflow in
 // the Airplane API.
 export async function __airplaneEntrypoint(params, workflowArgs) {
+  registerWorkflowRuntime();
```
colinking, line 13: Is this safe to call multiple times? e.g. that if the workflow runtime is registered, don't re-register.

IIRC, Temporal wants to re-use the V8 isolate eventually, in which case the entrypoint could get run multiple times.

pkg/build/workflow/worker-shim.js
```diff
@@ -1,7 +1,7 @@
 import { NativeConnection, Worker } from '@temporalio/worker';
 
 // Activity code runs in the same node process as the worker, so we import it here directly.
-import { createActivities } from "airplane/internal/runtime/workflow"
+import { createActivities } from "@airplane/workflow-runtime/internal";
```
colinking, line 4: Can you update our internal workflow testing guide w/ advice on testing changes to the workflow runtime?

pkg/build/node.go
```diff
@@ -282,8 +286,11 @@ func GenShimPackageJSON(pathPackageJSON string, isWorkflow bool) ([]byte, error)
 	}
 
 	if isWorkflow {
-		// airplane>=0.2.0-6 already includes Temporal as a dependency, and so we don't include it here.
-		pjson.Dependencies["airplane"] = "0.2.0-21"
+		// @airplane/workflow-runtime already includes the temporal dependencies, and so we don't need to include them
+		// here.
+		sdkVersion := "0.2.0-25"
+		pjson.Dependencies["airplane"] = sdkVersion
+		pjson.Dependencies["@airplane/workflow-runtime"] = sdkVersion
```
colinking, line 293: Talking out loud here... but IIUC, there are four cases:
1. user has both airplane + @airplane/workflow-runtime
2. user has airplane, shim has @airplane/workflow-runtime
3. user has @airplane/workflow-runtime, shim has airplane
4. shim has airplane + @airplane/workflow-runtime

1/4 should be straightforward. Since user dependencies are higher up in the directory structure than shim dependencies, 2 should work as long as the shim does not install its own `airplane`. As for 3, we could just not support that? If so, sgtm.

pkg/build/node.go
```diff
@@ -221,10 +221,14 @@ func node(root string, options KindOptions, buildArgs []string) (string, error)
...
 		
+		# npm >= 7 will automatically install peer dependencies, even if they're satisfied by the root. This is
+		# problematic because we need the @airplane/workflow-runtime package to register the workflow runtime in the
+		# runtime map that is utilized by the user's code, and so we explicitly request legacy behavior in this
+		# instance, which does not install peer dependencies by default.
 		RUN mkdir -p /airplane/.airplane && \
 			cd /airplane/.airplane && \
 			{{.InlineShimPackageJSON}} > package.json && \
-			npm install
+			npm install --legacy-peer-deps
```
colinking, line 231: I may be misreading this, but the docs mention this:

> UPDATE: npm versions 1, 2, and 7 will automatically install peerDependencies if they are not explicitly depended upon higher in the dependency tree. For npm versions 3 through 6, you will receive a warning that the peerDependency is not installed instead. (https://nodejs.org/uk/blog/npm/peer-dependencies/)

IIUC, won't it find the version of `airplane` higher up in the dependency tree? Any chance we're installing shim dependencies before user dependencies?

colinking (changes requested):
Just want to discuss the `--legacy-peer-deps` flag.

pkg/build/workflow/workflow-shim.js
calufornia: `registerRuntime` (and `registerWorkflowRuntime`) should both be idempotent, so as long as `registerWorkflowRuntime` is called at least once, workflow tasks should be supported. If this line is called multiple times, it will just replace the entry in the `runtimeMap` with an identical entry.

pkg/build/node.go
calufornia: That's what I would have assumed as well - however, in practice if we don't have this flag `airplane` still gets installed in `/airplane/.airplane/node_modules`, even if we move this command after the user code dependency installations. Further verified by exec-ing into the docker container, removing `/airplane/.airplane/node_modules`, and running `npm install` again - a copy of `airplane` is still produced, even though the parent directory has a version that satisfies `@airplane/workflow-runtime`'s peer dependencies.

pkg/build/node.go
calufornia: Yes exactly! 3 should ideally not happen (since they would need `airplane` to do anything with `@airplane/workflow-runtime`), and 1 is also unlikely to occur.

pkg/build/node.go
calufornia: I might be wrong here, but is it possible that npm is considering the two folders (`/airplane` and `/airplane/.airplane`) to have two different dependency trees? I'm wondering if the recursive check to see if a dependency is in some parent directory happens at runtime, not installation.

colinking: Does this get easier if we don't specify `airplane` as a peer dep in `@airplane/workflow-runtime`?

colinking (approved):

pkg/build/node.go
colinking: Also, does this work as-is with `yarn`?

calufornia: I think so - or moving it to be a devDependency 😛. From what I understand, npm won't install nested dev dependencies, and we only really need it to be a dependency at all when developing the Node SDK.

calufornia: Yep, this was tested with a `yarn` parent directory (`/airplane`)

calufornia: I guess the question becomes, do we prefer:
1. Including `airplane` as a dev dependency
2. Including `airplane` as a peer dependency and adding the `--legacy-peer-deps` flag
I'm personally in favor of the first option

colinking: Do we need it be a dependency at all though? 😄 If the compiler is yelling at you, you could use a dynamic import.

colinking: Oh actually, if this is just to make TS happy, then a dev dep might not be too bad...

leeweisberger (approved):

PR #220, opened 2022-08-09
Handle yarn workspace with a non-worskpace package

leeweisberger:
## Description

Before we were doing a shortcut where we assumed all packages were part of a workspace. Now we are using `yarn` commands to find workspace packages. Non-workspace packages will no longer be detected.

## Test plan

Wrote a unit test that failed before and now passes.


pkg/build/node_externals.go
```diff
@@ -116,17 +115,19 @@ func ListDependencies(pathPackageJSON string) ([]string, error) {
...
-	err := filepath.WalkDir(dir,
-		func(path string, di fs.DirEntry, err error) error {
-			if !strings.Contains(path, "node_modules") && !strings.Contains(path, ".airplane") &&
-				di.Name() == "package.json" {
-				pathPackageJSONs = append(pathPackageJSONs, path)
-			}
-			return nil
-		})
-	return pathPackageJSONs, err
+	workspaceInfo, err := getYarnWorkspaceInfo(rootPackageJSON)
```
colinking, line 120: 😍

colinking (approved):
This is much cleaner, thanks for improving it 🙌🏻

PR #221, opened 2022-08-10
Fix yarn workspace dependency installation

calufornia:
## Description

This PR contains a couple changes related to yarn workspaces that are required for https://github.com/airplanedev/node-sdk/pull/96 to work. In particular, when generating the shim `package.json`, look to see if any `package.json` between the root `package.json` and the script's `package.json` satisfies the `airplane` packages - if so, no need to install. Previously, we would only look at the root `package.json`; when using workspaces, dependencies are often declared in the workspace `package.json` file instead of the root, and so we should check there too.

## Test plan
Tested locally by deploying and running the tasks included in https://github.com/airplanedev/node-sdk/pull/96


pkg/build/node.go
```diff
@@ -238,13 +242,16 @@ func node(root string, options KindOptions, buildArgs []string) (string, error)
 
 		{{if .UsesWorkspaces}}
 		COPY . /airplane
+		WORKDIR /airplane
```
calufornia, line 245: I'm not sure if this is right, but my understanding is that `yarn/npm install` generally should not happen in the workspace directory itself, but instead at the root directory, so that yarn/npm can automatically flatten the dependency tree, link dependencies, etc.

calufornia: We don't need this - `yarn install` in a workspace works the same no matter where the command is run.

pkg/build/node.go
```diff
@@ -50,17 +50,16 @@ func node(root string, options KindOptions, buildArgs []string) (string, error)
 	if entrypoint == "" {
 		return "", errors.New("expected an entrypoint")
 	}
-	if err := fsx.AssertExistsAll(filepath.Join(root, entrypoint)); err != nil {
-		return "", err
-	}
```
leeweisberger, line 55: Why'd you remove this?

pkg/build/node_externals.go
```diff
@@ -20,23 +20,18 @@ var esmModules = map[string]bool{
...
-// This is used as a bit of a workaround for esbuild - we're using esbuild to transform code but
-// don't actually want it to bundle. We hit issues when it tries to bundle optional packages
-// (and the packages aren't installed) - for example, pg optionally depends on pg-native, and
-// using just pg causes esbuild to bundle pg which bundles pg-native, which errors.
-// TODO: replace this with a cleaner esbuild plugin that can mark dependencies as external:
-// https://github.com/evanw/esbuild/issues/619#issuecomment-751995294
-func ExternalPackages(rootPackageJSON string) ([]string, error) {
+// GetPackageJSONs list all the package.json files that belong to a workspace, or just the root package.json if not
+// using workspaces.
+func GetPackageJSONs(rootPackageJSON string) ([]string, bool, error) {
```
leeweisberger, line 25: What is this bool? Can you document it? I see that it's whether we use workspaces - just at least comment that

pkg/build/node_externals.go
```diff
@@ -77,6 +83,20 @@ func ExternalPackages(rootPackageJSON string) ([]string, error) {
 	return deps, nil
 }
 
+// ListDependenciesFromRoot lists all dependencies in a set `package.json` files.
```
leeweisberger, line 86: in a set of

pkg/build/node_externals.go
```diff
@@ -77,6 +83,20 @@ func ExternalPackages(rootPackageJSON string) ([]string, error) {
...
+func ListDependenciesFromRoot(packageJSONs []string) ([]string, error) {
```
leeweisberger, line 87: Not quite sure what `Root` means here. Maybe just ListDependencies?

leeweisberger (approved):
Makes sense to me.

leeweisberger:
> Makes sense to me.

Can you add a test that confirms that this works? Create a fixture with a workspace that includes a shim dependency somewhere in the workspace. Won't block, but would prefer if there was a test

colinking:
Thanks for the fix @calufornia! Will defer to @leeweisberger on the review.

PR #224, opened 2022-08-12
Expose ResourceFactory

calufornia:
## Description
In the CLI, we'd like a registry of all resource kinds so that we can print the available kinds to the user when they're setting a resource in the dev config file. This exports the resource factory map so that we can do that.

## Test plan
n/a


kellymtrinh (approved):

francescaleung:
I'd rather that we'd expose this in a more controlled way--e.g., `GetRegisteredResourceKinds`, if all you're looking for is keys. Exporting the whole map is pretty heavy-handed.

PR #230, opened 2022-08-17
Add kind configs to lib

calufornia:
## Description

This PR adds all the kind configs to the lib repo, which is necessary because we're changing the format of resources in the dev config file to use the api version of the resource, as opposed to the builtins versions of resources (i.e. `/pkg/resources/kinds`). We also include most of the interface methods of the kind configs since they give us the ability to convert the kind configs into a builtin resources (which we need for local dev), validation, etc.

## Test plan

n/a


kellymtrinh (approved):
cc @francescaleung

francescaleung (changes requested):
So, I have feelings about this, but am open to compromise. I don't think we should be exposing this outside of the API server / web. There are arguments for not having two different representations of resources, which I think is valid, but we're not there yet. I don't think this is better though.

calufornia:
> So, I have feelings about this, but am open to compromise. I don't think we should be exposing this outside of the API server / web. There are arguments for not having two different representations of resources, which I think is valid, but we're not there yet. I don't think this is better though.

Discussed offline, we'll move forward with landing this PR with the understanding that the api resource and external resource will eventually get merged and so there will be some tech debt in moving the previewer to use the external resource. I'll add a comment mentioning we want to remove this package from lib eventually.

francescaleung (approved):

PR #232, opened 2022-08-18
Add dependency installation step to support deploying config as code from worker

ameyakhare:
Resolves TSK-116

## Description
Adds a step to install dependencies (once per project root) during a deploy so that the worker can properly import user code to extract task and view configs.

## Test plan
Deploying locally works when I remove the node_modules directory and it doesn't reinstall if the directory exists.
Deploying locally is significantly faster now, despite having numerous files.


pkg/deploy/discover/parser/node/parser.js
```diff
@@ -0,0 +1,71 @@
+var path = require("node:path");
```
leeweisberger, line 1: Can you output a comment that this is autogenned please

pkg/deploy/discover/code_task_discover.go
```diff
@@ -142,6 +143,7 @@ func (c *CodeTaskDiscoverer) parseNodeDefinitions(ctx context.Context, file stri
 
 	parsedConfigs, err := extractConfigs(compiledJSPath)
 	if err != nil {
+		fmt.Println(err)
```
leeweisberger, line 146: You mean to print this?

pkg/deploy/discover/build.go
```diff
@@ -11,14 +12,52 @@ import (
...
 
+//go:generate npx -p typescript tsc parser/node/parser.ts
+//go:embed parser/node/parser.js
+var nodeParserScript []byte
+
 const outputDir = ".airplane"
 
+func maybeInstallDependencies(rootDir string) error {
+	if !fsx.Exists(path.Join(rootDir, "node_modules")) {
+		// node_modules doesn't exist, we may need to install dependencies
```
leeweisberger, line 30: What if their node_modules exists but is not up to date? Like I added a new dep in my code and package.json but didn't run npm install?

pkg/deploy/discover/build.go
```diff
@@ -11,14 +12,52 @@ import (
...
+				if isYarnBerry, _ := build.IsYarnBerry(packageJSONPath); isYarnBerry {
+					cmd = exec.Command("yarn", "install")
+				} else {
+					cmd = exec.Command("yarn", "install", "--non-interactive", "--production", "--frozen-lockfile")
+				}
+
+			} else if fsx.Exists(path.Join(rootDir, "package-lock.json")) {
+				cmd = exec.Command("npm", "ci", "--production")
+			} else {
+				cmd = exec.Command("npm", "install", "--production")
```
leeweisberger, line 46: Why are these different?

pkg/deploy/discover/code_task_discover.go
```diff
@@ -120,6 +117,10 @@ func (c *CodeTaskDiscoverer) parseNodeDefinitions(ctx context.Context, file stri
 		return nil, errors.Wrap(err, "unable to interpret task path metadata")
 	}
 
+	if err := maybeInstallDependencies(pm.RootDir); err != nil {
+		return nil, err
+	}
```
leeweisberger, line 122: So the theory is we should not be running this if the deploy is local. We will run it in the server. 

We are going to install deps twice in that case right? Once during discovery and then a second time during build. That's going to slow things down a bit.

Also we aren't caching deps here right?

leeweisberger (commented):
Want to make sure I understand correctly / we understand the perf implications here

pkg/deploy/discover/build.go
ameyakhare: if they are developing locally, we won't override their node_modules. unsure if that makes more sense or if we should always install when attempting to discover their tasks. I imagine the later option might be a little slow during local development and isn't something we do right now anyways

pkg/deploy/discover/build.go
ameyakhare: part of this has been copied from the build process (but not enough of it was similar enough to refactor and reuse)

pkg/deploy/discover/code_task_discover.go
ameyakhare: I believe npm/yarn caches dependencies in a global system cache. downloading a particular dependency would be slow until the cache warms up

pkg/deploy/discover/parser/node/parser.js
ameyakhare: I've removed this for now and will do it in another diff

leeweisberger (approved):
Approving, but please refactor the install logic out and reuse in node.go

Also add some more logging.

I'm still worried about adding so much latency installing dependencies twice without caching. We will probably have to come up with something better

pkg/deploy/discover/code_task_discover.go
leeweisberger: These worker pods are not very long lived, so we can't really count on that

ameyakhare:
I'm abandoning this in favor of the new deployments work instead

PR #234, opened 2022-08-19
Support node version and env vars in code-only config

ameyakhare:
## Description
Adds support for nodeVersion and envVars in code-only config.

## Test plan
Was able to deploy a task locally with these changes.


pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -44,8 +50,8 @@ const extractTaskConfigs = (files: string[]): AirplaneConfigs => {
   let taskConfigs: TaskDefWithBuildArgs[] = [];
   let viewConfigs: ViewDef[] = [];
   for (const file of files) {
-    const resolvedPath = path.relative(__dirname, file);
-    const lib = resolvedPath.replace(/.ts$/, "");
+    const relPath = path.relative(__dirname, file);
+    const lib = path.join(path.dirname(relPath), path.parse(relPath).name);
```
leeweisberger, line 54: Can you explain? this seems unrelated?

leeweisberger (changes requested):

pkg/deploy/discover/parser/node/parser.ts
ameyakhare: we actually dont need to do the path manipulation at all since we import from js files, so leaving the extension is fine

leeweisberger (approved):

PR #235, opened 2022-08-22
Stop prepending src to entrypoint by default

zhan-airplane:
## Description
Going to switch `airplane views dev` from using a temp dir to using a hidden `.airplane-view` dir in the original directory, hopefully will fix some of the strange temp dir issues we've been seeing. Also will make it easier to delete the dir if something goes wrong

## Test plan
None


stepzhou (changes requested):
This may be a dumb question but how does this PR change the temp directory to `.airplane-view`? I don't see it anywhere

stepzhou (approved):
Oh nvm this is part of a stacked PR

PR #236, opened 2022-08-22
Add resource conversion

calufornia:
## Description
Adds methods for converting from the external representation of resources to the internal representation which are used by the dev server API. In order to maintain the structure of the `resources/` directory, we choose not to add a `ToInternalResource` method to the external resource's interface, but instead create a `conversion/` directory that uses type switches to get the appropriate conversion method.

## Test plan
Tested in conjunction with https://github.com/airplanedev/cli/pull/588


pkg/resources/factory.go
```diff
@@ -53,3 +53,13 @@ func BaseFactory(serialized map[string]interface{}, result interface{}) error {
 	}
 	return nil
 }
+
+// RegisteredResourceKinds lists all registered resource kinds.
+func RegisteredResourceKinds() []ResourceKind {
+	kinds := make([]ResourceKind, 0, len(ResourceFactories))
```
francescaleung, line 59: at some point we should probably not export ResourceFactories

pkg/resources/conversion/rest.go
```diff
@@ -0,0 +1,57 @@
...
+		if len(authParts) != 2 {
+			return kind_configs.InternalResource{}, errors.New("authorization header is not of the form `<kind> <token>`")
+		}
+
+		switch authKind := authParts[0]; authKind {
+		case "Basic":
+			token := authParts[1]
+			credentials, err := base64.StdEncoding.DecodeString(token)
+			if err != nil {
+				return kind_configs.InternalResource{}, errors.Wrap(err, "decoding authorization token")
```
francescaleung, line 25: In this error case (& also line 30) it might be useful to log a warning somewhere & then just continue on instead of erroring out; someone could legitimately have a weird Authorization header in their config. Technically (as far as Airplane is concerned anyway) would not be an error case.

francescaleung (approved):

pkg/resources/factory.go
calufornia: Yep! this is the first step in doing that - once I switch CLI over to use this method, I'll remove (or at least unexport) the resource factory map.

PR #238, opened 2022-08-24
Simplify parser command

ameyakhare:
## Description
We don't need the typescript and types/node packages anymore when using tsx.

## Test plan
Still deploys properly.


colinking (approved):
Love to see it

PR #247, opened 2022-08-30
Align airplane package versions

calufornia:
## Description

If the user has `airplane` installed as a dependency, we should use the same version for the `@airplane/workflow-runtime` package - previously, the workflow-runtime package would just get set to the default SDK version.

## Test plan
Tested by deploying a workflow task and verifying that the SDK dependencies were in sync.


pkg/build/node_externals.go
```diff
@@ -85,52 +85,60 @@ func ExternalPackages(packageJSONs []string, usesWorkspaces bool) ([]string, err
...
-	return deps, nil
+	return deps, versions, nil
 }
 
-// ListDependencies lists all dependencies (including dev and optional) in a `package.json` file.
-func ListDependencies(pathPackageJSON string) ([]string, error) {
+// ListDependencies lists all dependencies (including dev and optional) and their versions in a `package.json` file. We
+// return the versions in a separate string slice instead of returning a single map in the case that the same dependency
+// is included in the production, dev, or optional dependencies.
+func ListDependencies(pathPackageJSON string) ([]string, []string, error) {
```
ameyakhare, line 106: maybe just return a map here to make parent code cleaner?

ameyakhare (approved):

PR #252, opened 2022-09-01
Add ScrubSensitiveData to Resource

francescaleung:
## Description

This change involves making all concrete Resource types pointers, so it happens to touch a lot of files. ScrubSensitiveData is mostly copied from the KindConfig implementation, but is modified to include calculated fields such as DSN & REST auth headers.

Resolves TSK-557.

## Test plan

Tested stdapi runs with relevant airport changes; tested previewer runs where resources are attached.


pkg/resources/kind_configs/redshift.go
```diff
@@ -41,5 +42,16 @@ func (this RedshiftKindConfig) dsn() string { // nolint:unused
 }
 
 func (this RedshiftKindConfig) ToExternalResource(base resources.BaseResource) (resources.Resource, error) {
-	return this.PostgresKindConfig.ToExternalResource(base)
+	res, err := this.PostgresKindConfig.ToExternalResource(base)
+	if err != nil {
+		return nil, err
+	}
+	postgres, ok := res.(*kinds.PostgresResource)
```
kellymtrinh, line 49: was there anything weird that came up with saving redshift configs? I forget if it's able to cast the type to both postgres and redshift here

pkg/resources/kinds/rest.go
```diff
@@ -75,7 +76,23 @@ func RESTResourceFactory(serialized map[string]interface{}) (resources.Resource,
...
-	return resource, nil
+	return &resource, nil
+}
+
+func (r *RESTResource) ScrubSensitiveData() {
+	scrubbedHeaders := map[string]string{}
+	for k, v := range r.Headers {
+		if isSecretHeader(r.SecretHeaders, k) {
+			scrubbedHeaders[k] = ""
+		} else {
```
kellymtrinh, line 87: Might not need the else case here

kellymtrinh (approved):

pkg/resources/kinds/rest.go
francescaleung: no, I think this does need to be here

PR #256, opened 2022-09-02
Add utility to generate inline code config js/ts tasks

ameyakhare:
## Description
Adds GenerateInline function that will generate an inline configured js/ts task definition. Added it to the runtime interface, so the implementation for the other languages is TBD.

## Test plan
Generated a file for this task: https://web.airstage.app/tasks/tsk20220701zjfvvatnlei?__env=prod
Output:
```
import airplane from "airplane"

export default airplane.task(
  {
    slug: "node_builtins_demo",
    name: "Node Builtins Demo",
    description: "This is a description.",
    parameters: {
      example_parameter: {
        type: "shorttext",
        name: "Example Parameter",
        description: "This is an example parameter.",
        default: "foobar",
        options: [
          "foobar",
          {label: "notegsdee", value: "eggsdee"},
        ]
      },
    },
    allowSelfApprovals: false,
    timeout: 3500,
    constraints: {
      mta: "v2",
    },
    resources: {
      db: "db",
      mail: "res_airplane_dev",
      website: "praygo",
    },
    nodeVersion: "16",
    envVars: {
      CONFIG_ENV_VAR: {
        config: "AIRPLANE_DATA_PLANE_BUCKET"
      },
      VALUE_ENV_VAR: "prayge",
    },
  },
  (params) => {
    // Put the main logic of the task in this function.
    console.log('parameters:', params);

    // You can return data to show outputs to users.
    // Outputs documentation: https://docs.airplane.dev/tasks/outputs
    return [
        {element: 'hydrogen', weight: 1.008},
        {element: 'helium', weight: 4.0026},
    ];
  }
)
```


francescaleung (changes requested):

pkg/runtime/javascript/javascript.go
```diff
@@ -47,6 +48,156 @@ export default async function(params) {
...
+var inlineCode = template.Must(template.New("ts").Funcs(template.FuncMap{
+	"toJSTypeVar":   ToJavascriptTypeVar,
+	"paramRequired": ParamRequired,
+}).Parse(`import airplane from "airplane"
+
+export default airplane.task(
+  {
+    slug: "{{.Slug}}",
+    {{- with .Name}}
+    name: "{{.}}",
```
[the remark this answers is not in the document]
ameyakhare: added!

pkg/runtime/javascript/javascript.go
```diff
@@ -49,45 +49,51 @@ export default async function(params) {
 `))
 
 // Inline code template function helpers.
-func ToJavascriptTypeVar(value interface{}) string {
+func escapeString(s string) string {
+	s = strings.ReplaceAll(s, "\"", "\\\"")
+	s = strings.ReplaceAll(s, "\n", "\\n")
+	return s
```
francescaleung, line 55: hmmm you almost want this to be json.Marshal without the first + last quotes. is that too much?

francescaleung (approved):

pkg/runtime/javascript/javascript.go
ameyakhare: I used json marshal now

PR #258, opened 2022-09-06
Rename index.js instead of changing package.json

leeweisberger:
Instead of changing the package.json to point at the dummy file, just rename index.es.js -> index.dummy.js. This ensures that all modules that import index.es.js (e.g. icons) import the dummy file

Note this relies on https://github.com/airplanedev/views/pull/247 to set the main field (previously we were doing this here)


leeweisberger:
Closing in favour of https://github.com/airplanedev/lib/pull/279

PR #273, opened 2022-09-09
Catch and set errors in python shim

ericliuche:
## Description
Catch python exceptions and display their errors like we do in node

## Test plan
Ran locally
Screen Shot 2022-09-09 at 9 40 42 AM


pkg/build/python-shim.py
```diff
@@ -44,8 +48,11 @@ def run(args):
                 print("Printing return values to stdout instead.", file=sys.stderr)
                 sys.stderr.flush()
                 print(json.dumps(ret, indent=2))
-    except Exception as e:
-        raise Exception("executing {{.Entrypoint}}") from e
+    except Exception:
+        tb = traceback.format_exc()
+        print(tb, file=sys.stderr)
+        airplane.set_output(tb, "error")
```
colinking, line 54: Can we print just the error message to output? And print the trace to the logs? That would mirror how we do it in Node. Ideally the error output is more readable for less-technical operators.

colinking (changes requested):
Otherwise LGTM.

pkg/build/python-shim.py
ericliuche: yup! sgtm

colinking (approved):

PR #274, opened 2022-09-09
Initialize an async function during inline init

colinking:
## Description

When initializing an inline code task, create an `async` function so that users can use `await` syntax.

## Test plan

Initialize an inline task locally.


colinking:
* main
  * **PR #274** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/274" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/274?utm_source=stack-comment).

ameyakhare (approved):

PR #275, opened 2022-09-12
Surface better error for malformed python tasks

ericliuche:
## Description
Surface a better error if a user is missing a `main` function or if the `main` function is malformed

## Test plan
ran locally, tested task with no params 

Screen Shot 2022-09-13 at 1 48 21 PM
Screen Shot 2022-09-13 at 1 48 33 PM


pkg/build/python-shim.py
```diff
@@ -34,6 +35,21 @@ def run(args):
...
+        if not hasattr(mod, "main"):
+            raise Exception("""Task is missing a `main` function. Add a main method like so and re-deploy:
+
+```
+def main(params):
+  print(params)
+```""")
+        num_params = len(inspect.signature(mod.main).parameters)
+        if num_params != 1:
+            raise Exception(f"""`main` function must have exactly 1 parameter, found {num_params}. Update the main method like so and re-deploy:
```
colinking, line 47: Do we not handle this? Because we should. If a task takes no parameters, you shouldn't need to define the `params` argument. This is how it works for Node.

So bit of scope-creep, but that should be a quick fix -- could you do that here?

pkg/build/python-shim.py
```diff
@@ -34,6 +35,21 @@ def run(args):
         {{if .EntrypointFunc}}
         ret = mod.{{.EntrypointFunc}}.base_func(json.loads(args[1]))
         {{else}}
+        if not hasattr(mod, "main"):
+            raise Exception("""Task is missing a `main` function. Add a main method like so and re-deploy:
```
colinking, line 39: nice

pkg/build/python-shim.py
```diff
@@ -34,6 +35,21 @@ def run(args):
         {{if .EntrypointFunc}}
         ret = mod.{{.EntrypointFunc}}.base_func(json.loads(args[1]))
         {{else}}
+        if not hasattr(mod, "main"):
```
colinking, line 38: This breaks with `"mod.{{.EntrypointFunc}}"` right?

pkg/build/python-shim.py
colinking, line 39: Does Python call them methods or functions? Let's be consistent.

```suggestion
            raise Exception("""Task is missing a `main` method. Add the following to your task and re-deploy:
```

colinking (changes requested):

pkg/build/python-shim.py
ericliuche: yup ill toss that in here

pkg/build/python-shim.py
ericliuche: yeah, but that's changing very soon when i add my inline code changes so ill just update it then

colinking: Any reason not to do it now? Since the shim already has inline config logic in it, I'd like to stay consistent with that.

ericliuche: making the change right now would break the current behavior, we need to first agree on the sdk change and then update the logic here. basically the semantics of how we're calling the user's inline code is changing from

`mod.{{.EntrypointFunc}}.base_func(json.loads(args[1]))`

to something like

`mod.{{.EntrypointFunc}}(**json.loads(args[1]))`

pkg/build/python-shim.py
ericliuche: ah @colinking, this doesn't work for node either... 

https://web.airstage.app/runs/run20220912z0m7sp6grw7?__env=prod

ericliuche: nvm nvm that was my bad, it does work

pkg/build/python-shim.py
```diff
@@ -31,10 +32,35 @@ def run(args):
...
+def main(params):
+  print(params)
+```
+"""
+        if not hasattr(mod, "main"):
+            raise Exception(f"""Task is missing a `main` function. Add a main function like so and re-deploy:
+{main_example}""")
+        num_params = len(inspect.signature(mod.main).parameters)
+        # If the task doesn't have any parameters
+        if not arg_dict:
```
colinking, line 50: This is checking if any args were passed to the task, not whether the task has any parameters.

Consider a task with a single optional parameter. If you don't pass a value, does this still work?

IOW, we should support the following:

- `main(params)` w/ `{}` -> kk
- `main()` w/ `{}` -> kk
- `main(params)` w/ `{"foo": "bar"}` -> kk
- `main()` w/ `{"foo": "bar"}` -> should error

colinking (approved):

pkg/build/python-shim.py
ericliuche: right now all the cases you listed are correct. with the single option parameter, if no value is passed both `main(params)` and `main()` work. but if a parameter is provided, only `main(params)` works

PR #279, opened 2022-09-12
Remove dummy

leeweisberger:
## Description

Remove need to shim airplane view exports. The reason why the views library can't be run in a node environment is because it has some code that assumes it's in a browser environment. We're going to find the same thing true with other packages that our users use. For a more generic fix, use the `jsdom` package in parser.js to apply a fake dom to the user's view - this should allow code that expects to be run in a browser to run successfully.

## Test plan

I wrote some unit tests for this. We're now actually testing real views code. This required me to push up the node_modules of the test views (otherwise we'd have to run yarn install on every fixture which is annoying. we could do it but it'd add a bunch of extra time and it works just pushing up node_modules).


ameyakhare (changes requested):
as discussed offline, please add a workflow step to install node modules in some common directory for the tests to run properly

pkg/deploy/discover/build.go
```diff
@@ -24,7 +22,7 @@ const outputDir = ".airplane"
 func esbuildUserFiles(rootDir string) error {
 	var files []string
 	err := filepath.WalkDir(rootDir, func(path string, d fs.DirEntry, err error) error {
-		if slices.Contains([]string{outputDir, "node_modules"}, d.Name()) {
+		if slices.Contains([]string{outputDir, "node_modules", ".airplane-view"}, d.Name()) {
```
leeweisberger, line 25: Unrelated change - this is a generated file that we should not need to build

pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -1,4 +1,7 @@
-const path = require("path");
+// After changing this file, run `yarn build` to build parser.js.
```
leeweisberger, line 1: idt it's worth the time adding this to presubmit.

pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -114,9 +117,11 @@ const extractTaskConfigs = (files: string[]): AirplaneConfigs => {
   return {
     taskConfigs,
     viewConfigs,
-  }
+  };
 };
 
+const dom = new JSDOM(`<!DOCTYPE html><body></div></body>`);
+(global as any).document = dom.window.document;
```
leeweisberger, line 124: Here's the magic. Set the document to the document generated by jsdom. This gives the code all document properties that a real browser has

ameyakhare (commented):
you need to
1) change the temp file to be a js file in extractConfigs
2) Not run tsx and instead run node directly on the parser file that you produce

ameyakhare:
additionally, @airplane/views is still not resolving properly in your tests

leeweisberger:
> additionally, @airplane/views is still not resolving properly in your tests

This relies on the "add main to package" pr that I merged in views but have not released yet

pkg/deploy/discover/build.go
ameyakhare, line 25: would be nice to extract this into a top-level var

ameyakhare (approved):
im assuming you understand whats happening for jsdom. I looked over it and it seems reasonable, but it would be nice to know the implications of using it (in a comment or something)

PR #280, opened 2022-09-12
Precalculate fields on REST auth when updating

francescaleung:
## Description

When updating a `RESTResource` from no auth to basic auth, headers weren't getting added to the resource. This fixes that by adding an explicit `Calculate()` function to the `Resource` interface & calling it at the end of every `Update()` implementation.

## Test plan

Tested locally.


francescaleung:
as discussed: Add `Calculate()` to Resource interface, include `Calculate()` call at the end of every `Update()` call.

kellymtrinh (approved):

PR #281, opened 2022-09-13
Prefer resource slug to resource name in SQL/REST tasks

francescaleung:
## Description

Resolves TSK-550.

## Test plan

Tested locally.


pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -124,7 +124,7 @@
               "type": "object",
               "properties": {
                 "resource": {
-                  "description": "The name of a database resource.",
+                  "description": "The slug of a database resource. (Deprecated: This can also be the name of the resource.)",
```
colinking, line 127:
```suggestion
                  "description": "The slug of a database resource. This also accepts resource names for legacy tasks, but that behavior is deprecated.",
```

francescaleung: I like the succinctness of the former; it also indicates the deprecation up front, which I think is useful. It will continue to work, probably forever, but everything else here nudges new people/tasks towards slugs as opposed to names.

PR #288, opened 2022-09-15
Add build options for preinstall/postinstall scripts

ericliuche:
## Description
Adds support for users to provide `airplane_postinstall.sh and `airplane_preinstall.sh` scripts to execute hooks in our build process. We look for these scripts in the user's tasks root directory. For node tasks, if users provide both a command in `airplane_postinstall.sh` and in their package.json, we use the latter.

## Test plan
added unit tests, ran locally with build hooks and verified they were being executed


pkg/examples/typescript/custompostinstall/yarn.lock
```diff
@@ -1,116 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
colinking, line 1: This is missing in the `installhooks` example.

pkg/build/python_test.go
```diff
@@ -38,6 +38,15 @@ func TestPythonBuilder(t *testing.T) {
 			},
 			SearchString: "[1]",
 		},
+		{
+			Root: "python/buildoptions",
```
colinking, line 42: rename this example too

pkg/build/node_test.go
```diff
@@ -177,7 +177,15 @@ func TestNodeBuilder(t *testing.T) {
...
-			Root: "typescript/custompostinstall",
+			Root: "typescript/installhooks",
+			Kind: TaskKindNode,
+			Options: KindOptions{
+				"shim":       "true",
+				"entrypoint": "main.ts",
+			},
+		},
+		{
+			Root: "typescript/buildoptions",
```
colinking, line 188: Let's make these examples consistent. Rename example to `installhooksviashell` and the other to `installhooksviapackagejson`?

pkg/build/python_test.go
```diff
@@ -38,6 +38,15 @@ func TestPythonBuilder(t *testing.T) {
...
+			Kind: TaskKindPython,
+			Options: KindOptions{
+				"shim":       "true",
+				"entrypoint": "main.py",
+			},
+			SearchString: "[1]",
```
colinking, line 48: Does this actually test preinstalling? It's looking for "[1]".

The postinstall example before would fail because `prettier` wouldn't get installed otherwise.

pkg/build/options.go
```diff
@@ -0,0 +1,47 @@
...
+
+// BuildOptions is a collection of build options across different task kinds.
+type BuildOptions struct {
+	PreInstallCommand  string
+	PostInstallCommand string
+}
+
+func GetBuildOptions(root string) (BuildOptions, error) {
+	var preInstallCommand string
+	preInstallPath := filepath.Join(root, preInstallScriptName)
```
colinking, line 24: Look for these scripts in every directory from the entrypoint's directory up to the root directory. Pick the one closest to the entrypoint. You can use our `fsx.Find` for this.

pkg/build/node_test.go
```diff
@@ -177,7 +177,15 @@ func TestNodeBuilder(t *testing.T) {
 			},
 		},
 		{
-			Root: "typescript/custompostinstall",
+			Root: "typescript/installhooks",
+			Kind: TaskKindNode,
+			Options: KindOptions{
+				"shim":       "true",
+				"entrypoint": "main.ts",
+			},
```
colinking, line 185: Same concern as the Python task -- this isn't actually testing that preinstall is running.

pkg/build/node.go
```diff
@@ -251,6 +266,10 @@ func node(root string, options KindOptions, buildArgs []string) (string, error)
 
 		{{.Args}}
 
+		{{if .PreInstallCommand}}
+		RUN {{.PreInstallCommand}}
```
colinking, line 270: This should execute the script itself, not the contents. Scripts will span multiple lines and include comments that won't work if you inline them into a `RUN` command. Hashbangs should also work.

Seems like you'll need to support both this + executing content directly for `package.json` commands.

pkg/build/node.go
```diff
@@ -32,12 +32,18 @@ type templateParams struct {
...
 	Args                             string
 }
 
 // node creates a dockerfile for Node (typescript/javascript).
-func node(root string, options KindOptions, buildArgs []string) (string, error) {
+func node(
+	root string,
+	options KindOptions,
+	buildArgs []string,
+	buildOptions BuildOptions,
```
colinking, line 45: This API is confusing to me, e.g. what is the diff between options/args/buildOptions? Can you clean it up?

colinking (changes requested):

pkg/build/options.go
colinking: Whoops, use `fsx.FindUntil` actually: https://github.com/airplanedev/lib/blob/b95ed2a190dd783650545f0476fdaf44ec857d66/pkg/utils/fsx/fsx.go#L61

pkg/build/python_test.go
ericliuche: Fixed so that the install hooks write files. The tasks read in the files and print their contents and we check the string is there

pkg/build/options.go
ericliuche: Done!

pkg/build/node.go
ericliuche: fixed

pkg/build/node.go
ericliuche: since we now need the task's entrypoint, moved the install hook login inside the `build/node.go` and `build/python.go` files.

pkg/build/install_hooks.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/airplanedev/lib/pkg/utils/fsx"
+	"github.com/pkg/errors"
+)
+
+const (
+	preInstallScriptName  = "airplane_preinstall.sh"
+	postInstallScriptName = "airplane_postinstall.sh"
+)
+
+type InstallHooks struct {
```
colinking, line 16: Can this be unexported? In general, we should not export fields whenever we can, as it keeps the API surface area smaller.

pkg/build/install_hooks.go
```diff
@@ -0,0 +1,56 @@
...
+	PostInstallFilePath string
+}
+
+// GetInstallHooks look for install scripts in every directory from the entrypoint's
+// directory up to the root directory.
+func GetInstallHooks(entrypoint string, root string) (InstallHooks, error) {
+	find := func(filename string) (string, error) {
+		if start, found := fsx.FindUntil(
+			// Entrypoint path is relative to the root so we need to combine them.
+			filepath.Join(root, filepath.Dir(entrypoint)),
```
colinking, line 28: Can `entrypoint` ever be a directory? If so, keep this in mind: https://github.com/airplanedev/cli/pull/659/files#diff-54c597b320ae143c8edd1055c0fb6d644d5e7e9031f790c373744f4bbdd155b4R85

pkg/build/install_hooks.go
```diff
@@ -0,0 +1,56 @@
...
+			filepath.Dir(root),
```
colinking, line 29: Same goes with `root`, though we may `filepath.Clean` this always in which case it's not an issue.

pkg/build/install_hooks.go
```diff
@@ -0,0 +1,56 @@
...
+	// Paths are relative to the task's root
+	PreInstallFilePath  string
+	PostInstallFilePath string
+}
+
+// GetInstallHooks look for install scripts in every directory from the entrypoint's
+// directory up to the root directory.
+func GetInstallHooks(entrypoint string, root string) (InstallHooks, error) {
+	find := func(filename string) (string, error) {
+		if start, found := fsx.FindUntil(
```
colinking, line 26: More idiomatic

```suggestion
		if dir, ok := fsx.FindUntil(
```

pkg/build/install_hooks.go
```diff
@@ -0,0 +1,56 @@
...
+		); found {
+			abspath := filepath.Join(start, filename)
+			if _, err := os.ReadFile(abspath); err != nil {
+				return "", errors.Wrapf(err, "install hooks: reading %s", abspath)
+			}
+			relpath, err := filepath.Rel(root, abspath)
+			if err != nil {
+				return "", errors.Wrapf(err, "install hooks: fetching relative path %s", abspath)
+			}
+			return relpath, err
```
colinking, line 40:
```suggestion
			return relpath, nil
```

pkg/build/install_hooks.go
```diff
@@ -0,0 +1,56 @@
...
+// GetInstallHooks look for install scripts in every directory from the entrypoint's
+// directory up to the root directory.
+func GetInstallHooks(entrypoint string, root string) (InstallHooks, error) {
+	find := func(filename string) (string, error) {
+		if start, found := fsx.FindUntil(
+			// Entrypoint path is relative to the root so we need to combine them.
+			filepath.Join(root, filepath.Dir(entrypoint)),
+			filepath.Dir(root),
+			filename,
+		); found {
```
colinking, line 31: It's idiomatic to keep the happy path unindented, and put error handling / sad path handling in these if clauses. 

```suggestion
		); !ok {
```

pkg/build/node.go
```diff
@@ -108,6 +109,11 @@ func node(
 	}
 	argsCommand := strings.Join(buildArgs, "\n")
 
+	installHooks, err := GetInstallHooks(entrypoint, root)
```
colinking, line 112: Minor, but should we just call these `hooks`? Since we might expose non-install hooks in the future.

Not important, since this is guessing about the future, but up to you.

pkg/build/node.go
```diff
@@ -268,6 +268,10 @@ func node(
 
 		{{if .PreInstallCommand}}
 		RUN {{.PreInstallCommand}}
+		{{else if .PreInstallPath}}
+		COPY {{ .PreInstallPath }} airplane_preinstall.sh
+		RUN chmod +x airplane_preinstall.sh
+		RUN ./airplane_preinstall.sh
```
colinking, line 274: This creates two layers. Let's merge:

```
RUN chmod +x airplane_preinstall.sh && ./airplane_preinstall.sh
```

pkg/build/node.go
```diff
@@ -278,6 +282,10 @@ func node(
 
 		{{if .PostInstallCommand}}
 		RUN {{.PostInstallCommand}}
+		{{else if .PostInstallPath}}
+		COPY {{ .PostInstallPath }} airplane_postinstall.sh
+		RUN chmod +x airplane_postinstall.sh
+		RUN ./airplane_postinstall.sh
```
colinking, line 288: same here

colinking (approved):
Thanks! LGTM.

pkg/build/install_hooks.go
ericliuche: realized we can just always assume entrypoint is a directory and it also works for a file, so just going to remove the Dir call

PR #290, opened 2022-09-15
Make definitions env-aware

francescaleung:
## Description

Adds env-awareness to definitions, some discoverers.

Supports TSK-40.

## Test plan

Tested with corresponding `cli` changes.


colinking (changes requested):
IMO, we should not be making discovery env-aware. Definitions should not vary from environment to environment.

IIUC: we need `ID <-> Slug` mappings in addition to a `Name -> ID` mapping from the default environment for SQL/REST for backwards compatibility.

I'd rather we expose a `ListResourcesMetadata` API that returns resource metadata along with a nullable `resource` from the default environment. This would allow you to retain the above behavior without making discovery env-aware.

francescaleung:
The reason this came up is that someone had different resources defined in different environments; i.e., they had `foo` defined in their default environment & `bar` defined in their dev environment & wanted to deploy to their dev environment using `bar`. It was failing, because we pull resources from the default environment only.

Should we encourage folks to have the same definition file per environment? Probably. Should we _dictate_ that? I don't know.

Though I like the workaround for returning the resource in the default environment for `ListResourceMetadata`, which preserves existing by-name behaviour but allows for the above scenario as well.

francescaleung:
closing in favour of ListResourceMetadata

PR #295, opened 2022-09-21
Add GetResourceWithCredentialsResponse

calufornia:
## Description

Adds a `GetResourceWithCredentialsResponse` struct to libapi, which will be used by the CLI's client. Also adds an `envSlug` argument to the `ListResources` method of the API client interface. This method isn't used anywhere, and so this is a safe change.

## Test plan
n/a


pkg/api/client.go
```diff
@@ -119,6 +119,10 @@ type ListResourcesResponse struct {
 	Resources []Resource `json:"resources"`
 }
 
+type GetResourceWithCredentialsResponse struct {
```
colinking, line 122: Do you not want to add `GetResourceWIthCredentials` as a method to the API client?

colinking (approved):

pkg/api/client.go
calufornia: It looks like those methods are required if they're called inside lib - I notice that we just use this file to store general request/response structs, even if they're only used in CLI.

PR #299, opened 2022-09-23
Add RawCredentials to BigQuery

francescaleung:
## Description

Problem: BigQuery tests on new resources don't work because the base64-encoding of credentials only happens on Update(). This PR tracks RawCredentials separately from Credentials so that one is purely calculated & t'other is not; it looks for RawCredentials on updates/calculations first, & falls back to Credentials (as raw credentials).

## Test plan

Tested with & without web changes. `lib` will get updated in `airport` and `cli` before `web` gets updated.


kellymtrinh (approved):
🥲🙌

PR #301, opened 2022-09-23
Move builtins local client from cli to lib

kellymtrinh:
## Description
This helps support local SQL and REST execution. Breaking this out into a smaller PR.

Moves https://github.com/airplanedev/cli/blob/main/pkg/dev/builtins.go from `cli` to lib.
The builtin client is used in the sql and rest `prepareRun` implementations (also in lib) so I thought it made sense to move it here.

Goes with: https://github.com/airplanedev/cli/pull/682/files

## Test plan
- [x] delete the `builtins.go` package from cli and import it from this version of lib. Verify that builtin execution works in the editor.


calufornia (approved):
nit: You could probably just move `local_client.go` directly under `pkg/builtins/` instead of creating a subpackage, that way you could create a local client by doing something like `builtins.NewLocalClient` instead of `local_client.NewClient`

PR #302, opened 2022-09-24
Implement local SQL and REST runtimes

kellymtrinh:
## Description

- Implements the local SQL and REST runtimes so users can run SQL/REST tasks using `airplane dev` or in the editor. The main logic is in each type's `PrepareRun` function, used in CLI. 

- The `pkg/expressions` folder is pulled from airport (minus the v8go part, which requires CGO 😓 ). I will followup with a PR to pull this into airport. This is to support the common sql/rest use cases of using params or configs in JSTs: `value: {{params.my_value}}` and `{{configs.API_KEY}}` . 

## Test plan

 <details><summary> test child REST and SQL tasks from `airplane.execute()` 
</summary>

Screen Shot 2022-09-27 at 3 53 01 PM
(I checked that the return values of the sql and rest subtasks was expected)
</details>

<details><summary> test REST task execution works, with both config and param JSTs </summary>
Rest yaml, that uses both params and config JSTS: 
Rest yaml:

```
name: rest test
slug: rest_test
parameters:
  - name: User name
    slug: user_name
    type: shorttext
    description: Username
  - name: file name
    slug: my_file
    type: shorttext
    description: file
rest:
  resource: my_rest
  method: GET
  urlParams:
    email: "{{params.user_name}}"
    file: "{{params.my_file}}"
  path: /posts/1
  bodyType: json
  body: "{{params.my_file}}"
  formData:
    file: "{{params.my_file}}"
  headers:
    X-API-KEY: "{{configs.API_KEY}}"
    X-API-KEY-2: "{{params.my_file}}"
```

I verified the param values are correct, and the config comes from the config set in the `dev.yaml`:
Screen Shot 2022-09-27 at 3 57 22 PM

</details>

<details><summary>test SQL execution, with JST queryArgs</summary>

Screen Shot 2022-09-27 at 3 49 40 PM

SQL yaml:
```
name: test sql
slug: test_sql
description: Selects all runs from a team with x limit. Team ID example is 1jq9NDeTcTmyqqXAjEsk7V0wXaV
parameters:
- name: limit
  slug: limit
  type: float
- name: team id
  slug: name
  type: shorttext
sql:
  resource: my_db
  entrypoint: test_sql.sql
  queryArgs:
    user_limit: "{{params.limit}}"
    team_id: "{{params.name}}"
  configs:
  - test_me
```

sql file:
```
select * from runs where team_id=:team_id limit :user_limit;
```

</details>

 <details><summary> test regular non editor `airplane dev` for REST task and SQL task:
</summary>

Screen Shot 2022-09-27 at 4 21 07 PM
Screen Shot 2022-09-27 at 4 21 50 PM

</details>


pkg/build/versions.go
```diff
@@ -7,19 +7,19 @@ import (
 	"github.com/pkg/errors"
 )
 
-//go:embed versions.json
```
kellymtrinh, line 10: CI was complaining this file wasn't go fmted :/

pkg/expressions/template_test.go
```diff
@@ -0,0 +1,316 @@
...
+		{
+			// handles templates that are just an expression
+			Template: `{{ block1.str }}`,
+			Output:   `world`,
+		},
+		{
+			// handles tabs inside of expressions
+			Template: `{{	block1.str	}}`,
+			Output:   `world`,
+		},
```
kellymtrinh, line 145: Screen Shot 2022-09-27 at 3 29 23 PM

I'm still debugging why this is complaining about this file -- running both gofmt or golangci-lint locally does not say anything but I'm going to try to use the exact version. :/ This is directly from airport

kellymtrinh: Resolved by 
```
 docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:v1.47.2 golangci-lint run --fix -v --timeout 5m
```

pkg/runtime/eval_params.go
```diff
@@ -0,0 +1,34 @@
...
+func EvalParams(ctx context.Context, inputs, paramValues, configs map[string]interface{}) (map[string]interface{}, error) {
+	params := map[string]interface{}{}
+	c := expressions.NewLookupClient()
+	for k, arg := range inputs {
+		argString, ok := arg.(string)
+		if !ok {
+			return nil, fmt.Errorf("invalid input: '%v' for %s", arg, k)
+		}
+		tmpl := expressions.NewTemplate(argString)
+		result, err := tmpl.Evaluate(ctx, c, expressionsiface.EvaluateOpts{
```
calufornia, line 23: Should we handle the error here?

pkg/runtime/eval_params_test.go
```diff
@@ -0,0 +1,34 @@
+package runtime
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestEvalParams(t *testing.T) {
```
calufornia, line 10: Also test when there's an expression that can't be evaluated (i.e. isn't in `paramValues` or `configValues`)

pkg/runtime/sql/sql.go
```diff
@@ -31,7 +33,34 @@ type Runtime struct{}
...
+		queryArgs, ok := opts.KindOptions["queryArgs"].(map[string]interface{})
+		if !ok {
+			return nil, nil, errors.New("invalid queryArgs")
+		}
+		evalOpts, err := runtime.EvalParams(ctx, queryArgs, opts.ParamValues, opts.ConfigVars)
+		if err != nil {
+			logger.Warning(err.Error())
+		}
+		opts.KindOptions["queryArgs"] = evalOpts
+
```
calufornia, line 47:
```suggestion
```

pkg/runtime/sql/sql.go
```diff
@@ -31,7 +33,34 @@ type Runtime struct{}
...
-	return nil, nil, runtime.ErrNotImplemented
+
+	if opts.KindOptions != nil && opts.KindOptions["queryArgs"] != nil {
+		queryArgs, ok := opts.KindOptions["queryArgs"].(map[string]interface{})
+		if !ok {
+			return nil, nil, errors.New("invalid queryArgs")
+		}
+		evalOpts, err := runtime.EvalParams(ctx, queryArgs, opts.ParamValues, opts.ConfigVars)
+		if err != nil {
+			logger.Warning(err.Error())
```
calufornia, line 44: How come we don't return here like above?

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1216,6 +1216,12 @@ func (d *Definition_0_3) GetConfigAttachments() ([]api.ConfigAttachment, error)
 }
 
 func (d *Definition_0_3) GetResourceAttachments() map[string]string {
+	if d.SQL != nil {
```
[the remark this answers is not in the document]
kellymtrinh: Updated with a comment- sql and rest definitions don't store their resource attachments on the same field as other definitions. They also don't have aliases, so this was returning empty before when it should have returned whatever was attached

pkg/runtime/sql/sql.go
kellymtrinh: At first I thought it might be better to do best evaluation and not error if the JST doesn't work (since it's not exactly the same as prod), but I updated to return the error instead. If it's not a JST it doesn't get evaluated, which is expected anyway.

pkg/runtime/eval_params.go
```diff
@@ -0,0 +1,45 @@
...
+type EvalGlobals struct {
+	ParamValues map[string]interface{}
+	Configs     map[string]interface{}
+}
+
+// EvalTemplates takes in the user input and evaluates the JST against
+// configs and param values. Example: {{configs.API_KEY}} and {{params.user_name}}
+// Uses the expressions package until we move to V8
+func EvalTemplates(ctx context.Context, inputs map[string]interface{}, globals EvalGlobals) (map[string]interface{}, error) {
+	params := map[string]interface{}{}
```
calufornia, line 21: nit: Maybe call this something different to differentiate with the param globals

pkg/runtime/eval_params.go
```diff
@@ -0,0 +1,45 @@
...
+			Globals: map[string]expressionsiface.ExpressionsLookupFn{
+				"params":  expressions.LookupFn(globals.ParamValues),
+				"configs": expressions.LookupFn(globals.Configs),
+			},
+		})
+		if err != nil {
+			return nil, errors.Wrap(err, "error evaluating template")
+		}
+		if result.Output == nil {
+			params[k] = ""
```
calufornia, line 39: Is setting this to the empty string safe? What if param is supposed to be a non-string value?

pkg/runtime/rest/rest.go
```diff
@@ -0,0 +1,109 @@
...
+	return filepath.Dir(path), nil
+}
+
+// Kind implementation.
+func (r Runtime) Kind() build.TaskKind {
+	return build.TaskKindREST
+}
+
+// FormatComment implementation.
+func (r Runtime) FormatComment(s string) string {
```
calufornia, line 96: Does the `FormatComment` method apply to REST tasks? REST tasks don't have an associated file right?

pkg/runtime/rest/rest.go
```diff
@@ -0,0 +1,109 @@
...
+	// will fallback to the task kind
+	runtime.Register(".rest", Runtime{})
+}
+
+// Runtime implementation.
+type Runtime struct{}
+
+// PrepareRun implementation.
+func (r Runtime) PrepareRun(ctx context.Context, logger logger.Logger, opts runtime.PrepareRunOptions) (rexprs []string, rcloser io.Closer, rerr error) {
+	client, err := builtins.NewLocalClient(goruntime.GOOS, goruntime.GOARCH, logger)
```
calufornia, line 33: It seems like a lot of this code is duplicated between rest and sql - can we extract this out into a helper?

calufornia (approved):

pkg/runtime/eval_params.go
kellymtrinh: If not, the default behavior from the evaluator is different for:
```
"{{configs.doesnt_exist}}" //defaults to nil
"{{doesnt_exist.whatever}}" // defaults to "" empty string	 
```

Took it out in case it's confusing, but this should still be ok for letting the user know their JST is giving back something invalid.

pkg/runtime/rest/rest.go
kellymtrinh: They don't - updated to have a comment. It needs to have this because it's in the runtime interface.

PR #304, opened 2022-09-26
Load envSlug from get param in main.tsx if it exists

zhan-airplane:
## Description
Otherwise if someone promotes a view between environments, we don't have a way of indicating to the view that the env is updated (since the env slug is baked into the view)

## Test plan
Ran locally

---
cc @stepzhou @rose-huang @geodn


pkg/build/views/main.tsx
```diff
@@ -8,11 +8,13 @@ import React from "react";
 import ReactDOM from "react-dom/client";
 import App from "{{.Entrypoint}}";
 
+const paramsEnvSlug = (document) ? (new URL(document.location)).searchParams.get("envSlug") : null;
```
leeweisberger, line 11:
```suggestion
const paramsEnvSlug = typeof window !== 'undefined' ? (new URL(document.location)).searchParams.get("envSlug") : null;
```

leeweisberger, line 11: I don't think this is going to work in an iframe. I think you can test this in the editor where we are also rendering the view in an iframe. I **think** this is going to return to the location of the iframe which is different than the location in the parent browser.

Let me know if you test this and I'm wrong, but I'm pretty sure it won't work.

We are passing in the query params as url fragments. See the attached photo. I believe that's the url you're gonna get with window.location

image

You can read in the env using the fragment. But this also has to work if this isn't in an iframe. See lib/src/routing/useRouter.ts in views for how we grab the search params regardless of whether the view is in an iframe.

leeweisberger (changes requested):

pkg/build/views/main.tsx
leeweisberger: ^ that is safe in a non-web environment

leeweisberger: I saw your other pr. Ok now I get why this works when you tested it :)

leeweisberger (approved):

pkg/build/views/main.tsx
zhan-airplane: Yeah haha I *do* want the iframe location, not the parent browser location

zhan-airplane:
BTW gonna change from `envSlug` to `__env` just for consistency with other parts of web code

PR #305, opened 2022-09-26
Fix update handling for BigQuery credentials

francescaleung:
## Description

Problem: updates weren't being handled, due to trying to backfill RawCredentials. Fixed, now with unit tests (since I've gotten this wrong now twice).

## Test plan

Now with unit tests.


francescaleung:
cc @kellymtrinh

pkg/resources/kinds/bigquery_test.go
```diff
@@ -0,0 +1,84 @@
...
+				Location:    "location",
+				DataSet:     "dataset",
+				Credentials: credentials1,
+			},
+			update: &BigQueryResource{
+				ProjectID: "projectID",
+				Location:  "location",
+				DataSet:   "dataset",
+			},
+			expectedCredentials: credentials1,
```
calufornia, line 36: Is it possible to also verify `expectedRawCredentials: rawCredentials1` here?

calufornia (approved):

PR #306, opened 2022-09-26
Replace default scaffolding for JS, TS, and PY tasks

colinking:
## Description

Updates the scaffolding used when initializing tasks to include a more meaningful example.

Resolves TSK-707 -- I'll update the docs after this merges.

## Test plan

Initialized and locally executed a scaffolded task of each kind.


colinking:
* main
  * **PR #306** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/306" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/306?utm_source=stack-comment).

leeweisberger (approved):
It is semi worrisome that this did not cause a test to fail :( I guess we don't have tests. it would be cool if we did

PR #313, opened 2022-09-28
Fix BigQuery validation

francescaleung:
## Description

Fixes error that prevents SQL blocks with BigQuery resources from running.

## Test plan

Ran BigQuery block locally, created new BigQuery resource, edited existing BigQuery resource.


francescaleung:
cc @calufornia

colinking (approved):

PR #309, opened 2022-09-27
Remove error name from JS error messages

colinking:
## Description

Before:

CleanShot 2022-09-27 at 12 37 47

After:
CleanShot 2022-09-27 at 12 40 00

Relates to TSK-452 -- future PRs will improve these error messages.

## Test plan

Ran a task and a workflow and confirmed they both used the new error format.

PR #314, opened 2022-09-28
Log workflow errors as info not debug

colinking:
## Description

Marks caught errors and their output as info (rather than debug) outputs, matching the behavior of standard tasks.

Also fixes a bug in https://github.com/airplanedev/lib/pull/309 where `message` was unused.

Resolves TSK-784

## Test plan

Ran a workflow locally and confirmed that the output showed as expected.


colinking:
* main
  * **PR #314** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/314" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/314?utm_source=stack-comment).

calufornia (approved):

PR #318, opened 2022-09-29
Improve docker build caching when using node workspaces

yolken-airplane:
## Description
This change updates our builder node Dockerfiles so that we copy only the `package*.json` and `yarn.*` configs before running `yarn install` or `npm ci`. Then, after that's done, we copy the rest of the code.

This should improve caching in the workspaces case because code-only changes shouldn't require re-pulling all of the dependencies.

Resolves TSK-590

cc: @leeweisberger 

## Test plan
Testing completed successfully via:

1. Existing unit tests
2. New unit test for the added functionality
3. Building an example from the js-sdk repo (that uses workspaces) and verifying that the automated behavior matched the manual fixes suggested in [this doc](https://airplane.slab.com/posts/workflows-local-testing-w-js-sdk-b47iq73a#hn3v3-yarn-workspace-caching):

```
...
[ 8/15] COPY examples/standard/package*.json examples/standard/yarn.* /airplane/examples/standard/
[ 9/15] COPY examples/workflow/package*.json examples/workflow/yarn.* /airplane/examples/workflow/
[10/15] COPY package*.json yarn.* /airplane/
[11/15] COPY packages/airplane/package*.json packages/airplane/yarn.* /airplane/packages/airplane/
[12/15] COPY packages/workflow-runtime/package*.json packages/workflow-runtime/yarn.* /airplane/packages/workflow-runtime/
[13/15] RUN yarn install --non-interactive --production --frozen-lockfile && yarn cache clean
[14/15] COPY . /airplane
...
```


pkg/build/node.go
```diff
@@ -134,6 +135,13 @@ func node(
 		return "", err
 	}
 
+	if usesWorkspaces {
```
colinking, line 138: This is unrelated to this PR's changes -- @leeweisberger: do you know why `templateParams.UsesWorkspaces` isn't set to `usesWorkspaces`? Can the two value diverge?

pkg/build/node.go
```diff
@@ -134,6 +135,13 @@ func node(
...
+		cfg.WorkspacePackageCopyCmds, err = GetWorkspacePackageCopyCmds(root, packageJSONs)
```
colinking, line 139: This specifically excludes the root package.json, correct? Since we already add a `COPY` command for that directly in the `Dockerfile`. Maybe we rename the variable to `workspacePackageJSONs`? Or refactor a bit so that this instead generates all the `COPY` commands we need, regardless of whether workspaces are used?

colinking (approved):
Nice!

pkg/build/node.go
yolken-airplane: Oh, actually, it looks like it does include the root one. Let me fix that...

yolken-airplane: Ok, cleaned this up so that we can use the same logic whether workspaces are used or not. @colinking can you take another look?

leeweisberger (approved):

PR #319, opened 2022-09-30
Normalize definition in discoverer

francescaleung:
## Description

I looked into adding this as a necessary step of `Unmarshal`, but `Unmarshal` is called (via `taskdir.ReadDefinition`) in so many places that don't actually need normalization (read the definition to get the slug, read the definition to get the entrypoint) that it would have been way more invasive to add it to `Unmarshal`. It would have also added a ton more API requests to a lot of commands, & (to my knowledge) we don't have any caching in the APIClient. (Maybe we should?) I added it as an optional step instead, which means that there is still handling for `Resource` being the name of a resource rather than a slug in other places, since there's no guarantee that the caller has called `Normalize` before calling (say) `GetUpdateTaskRequest`.

Resolves TSK-735.

## Test plan

Tested locally.


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -620,6 +644,30 @@ func (d *RESTDefinition_0_3) getConfigAttachments() []api.ConfigAttachment {
...
+	if err != nil {
+		return err
+	}
+
+	// Check slugs first.
+	if _, ok := collection.bySlug[d.Resource]; ok {
+		return nil
+	} else if id, ok := collection.byName[d.Resource]; ok {
+		if d.Resource, ok = slugsByID[id]; !ok {
+			return errors.Errorf("missing resource slug: %s", id)
```
colinking, line 663: Clarify this error message. Same with REST.

```suggestion
			return errors.Errorf("could not find slug for resource with id: %q", id)
```

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -620,6 +644,30 @@ func (d *RESTDefinition_0_3) getConfigAttachments() []api.ConfigAttachment {
...
+		}
+	} else {
+		return errors.Errorf("unknown resource: %s", d.Resource)
```
colinking, line 666: Use the new `ResourceMissingError` here and with REST.

PR #321, opened 2022-10-03
Add GraphQL resource kind

francescaleung:
## Description

Resolves TSK-491.

## Test plan

Untested so far, but also unused.


pkg/resources/kinds/graphql.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+var ResourceKindGraphQL resources.ResourceKind = "graphql"
+
+func init() {
+	resources.RegisterBaseResourceFactory(ResourceKindGraphQL, func() resources.Resource { return &GraphQLResource{} })
+}
+
+type GraphQLResource struct {
+	resources.BaseResource `mapstructure:",squash" yaml:",inline"`
```
kellymtrinh, line 15: Does the RESTResource already include the BaseResource?

kellymtrinh (approved):

pkg/resources/kinds/graphql.go
francescaleung: It does, but we're using this pattern in RedshiftResource as well, so there's precedent for this already.

PR #322, opened 2022-10-03
Add support for inline python init

ericliuche:
## Description
Add support for inline python init from task

## Test plan
ran locally, will add tests


ameyakhare (approved):
nice tests

PR #323, opened 2022-10-03
Disable warning from importing Node modules

colinking:
## Description

Disables the Temporal warning when a workflow imports Node modules.

Ideally we keep this error around and perform additional tree-shaking to remove unused code, but Temporal's current webpack setup does not do any tree-shaking.

Resolves TSK-671

cc @yolken-airplane @ericliuche 

## Test plan

Deploy a workflow that attempts to import a Node module, e.g. a workflow with the following in it:

```typescript
import * as fs from 'fs'

export function foobar() {
  const exists = fs.existsSync("./main.airplane.ts")
  return exists
}

// the rest of the workflow...
```


colinking:
* main
  * **PR #323** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/323" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/323?utm_source=stack-comment).

pkg/build/workflow/workflow-bundler.js
```diff
@@ -5,15 +5,30 @@ const { writeFile } = require('fs/promises');
...
+        // case, as long as those packages are not imported from the workflow, we should
+        // not error.
+        externals: undefined,
+      }
+    },
   });
 
-  await Promise.all([
-    writeFile('/airplane/.airplane/workflow-bundle.js', code),
-    writeFile('/airplane/.airplane/workflow-bundle.map.js', sourceMap),
```
colinking, line 15: Source maps are returned inline nowadays, so this is unused.

calufornia (approved):

PR #324, opened 2022-10-03
Produce warning if using a Node built-in module

colinking:


colinking:
* main
  * **PR #323** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/323" target="_blank">Graphite</a> 
    * **PR #324** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/324" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/324?utm_source=stack-comment).

pkg/build/workflow/workflow-bundler.js
```diff
@@ -1,14 +1,55 @@
...
+    throw getError(\`${moduleName}.\${property}\`)
+  },
+  // Handles applying the default module export as a function
+  apply: () => { throw getError(); },
+  // Handles calling a constructor on the default module export
+  construct: () => { throw getError(); },
+});
+
+module.exports = proxy;`;
+    fallbacks[moduleName] = `/airplane/.airplane/shims-${moduleName.replace(/\//g, "-")}.js`;;
```
calufornia, line 40:
```suggestion
    fallbacks[moduleName] = `/airplane/.airplane/shims-${moduleName.replace(/\//g, "-")}.js`;
```

pkg/build/workflow/workflow-bundler.js
```diff
@@ -1,14 +1,55 @@
...
 // for details.
 async function bundle() {
+  // Generate a fallback for each of the built-in Node modules.
+  const fallbacks = {}
+  await Promise.all(shimmedModules.map(async (moduleName) => {
+    const shim = `const getError = (path = "${moduleName}") => {
+  return new Error(\`Workflows do not have access to Node.js built-in packages (cannot access "\${path}"). Move that logic into a Node.js task and call the task from your workflow.\`);
+}
+
+function proxyTarget() {
```
calufornia, line 19: Just so I understand, this is a dummy function, right? i.e. it doesn't matter what the contents of this function are, as long as we export an object whose properties we've trapped?

calufornia (approved):
🔥

pkg/build/workflow/workflow-bundler.js
colinking: The contents can be anything, but I choose a function since functions can be constructed and applied, which allows us to proxy those methods, too, besides `get`. The underlying function won't get called, but I added the error (which I just fixed so it uses the same error as the rest of the proxy -- thanks for catching) in case.

PR #325, opened 2022-10-03
Add yarn workspace caching (second attempt)

yolken-airplane:
## Description
This change is a retry of https://github.com/airplanedev/lib/pull/318 but without the improved caching behavior if there are any pre/post-install hooks in the user code or a provided airplane preinstall script. The problem with the latter is that they might assume that the entire code is present, which can cause the build to fail.

cc: @leeweisberger 

## Test plan
Testing completed successfully via adding a new unit tests to cover the hook behavior. I've verified that the test fails if I remove the hook checks but passes with them.


pkg/build/node.go
```diff
@@ -134,6 +136,32 @@ func node(
...
+	if cfg.HasPackageJSON {
+		cfg.PackageCopyCmds, err = GetPackageCopyCmds(root, packageJSONs)
+		if err != nil {
+			return "", err
+		}
+
+		// Check all files for pre- or post-install scripts. If there are any found, then
+		// we to run the install with the entire codebase to be safe as opposed to
+		// just the package.json and yarn files (since the postinstall scripts might assume
+		// that all code is present).
```
colinking, line 150:
```suggestion
		// that all code is present). This prevents us from caching the user's dependencies separately from
		// their code.
```

pkg/build/node.go
```diff
@@ -254,14 +292,12 @@ func node(
 			{{.InlineShimPackageJSON}} > package.json && \
 			npm install --legacy-peer-deps
 
-		{{if .HasPackageJSON}}
-		COPY package*.json yarn.* /airplane/
+		{{if .InstallRequiresCode}}
+		COPY . /airplane
 		{{else}}
-		RUN echo '{}' > /airplane/package.json
+		{{range .PackageCopyCmds}}
```
colinking, line 298: Move this out of the `else`; it should always run since you could have an airplane pre-install script yet not `packag.json`

pkg/build/node_externals.go
```diff
@@ -171,6 +223,29 @@ func hasWorkspaces(pathPackageJSON string) (bool, error) {
...
 
+func hasInstallHooks(pathPackageJSON string) (bool, error) {
+	if _, err := os.Stat(pathPackageJSON); errors.Is(err, os.ErrNotExist) {
+		return false, nil
+	}
+	var pkg pkgJSON
+	buf, err := os.ReadFile(pathPackageJSON)
+	if err != nil {
+		return false, errors.Wrapf(err, "node: reading %s", pathPackageJSON)
+	}
```
colinking, line 234: Minor, but better to handle the `ErrNotExist` from `ReadFile` since the file could _technically_ get removed in-between.

pkg/build/node_externals.go
```diff
@@ -171,6 +223,29 @@ func hasWorkspaces(pathPackageJSON string) (bool, error) {
...
+	if err := json.Unmarshal(buf, &pkg); err != nil {
+		return false, errors.Wrapf(err, "parsing %s", pathPackageJSON)
+	}
+
+	if _, ok := pkg.Scripts["postinstall"]; ok {
+		return true, nil
+	}
+	if _, ok := pkg.Scripts["preinstall"]; ok {
+		return true, nil
+	}
```
colinking, line 245: There are a few other npm scripts: https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts

Let's also return `true` for prepare, prepublish, and dependencies.

pkg/examples/typescript/yarnworkspacespostinstall/pkg1/src/index.ts
```diff
@@ -0,0 +1 @@
+export const name = "1"
```
colinking, line 1: You'll want to confirm that the `postinstall` hook is actually running. Put something unique here and then add it as a check above.

```suggestion
export const name = "Look! Up in the sky! It's a bird! It's a plane! It's Superman!"
```

pkg/build/node_test.go
```diff
@@ -148,6 +148,15 @@ func TestNodeBuilder(t *testing.T) {
...
 			},
 		},
+		{
+			Root: "typescript/yarnworkspacespostinstall",
+			Kind: TaskKindNode,
+			Options: KindOptions{
+				"shim":       "true",
+				"entrypoint": "pkg2/src/index.ts",
+				"workdir":    examples.Path(t, "typescript/yarnworkspaces/pkg2"),
+			},
```
colinking, line 158: see other comment

```suggestion
			},
			SearchString: "Look! Up in the sky! It's a bird! It's a plane! It's Superman!"
```

colinking (approved):

pkg/build/node.go
leeweisberger: Well wouldn't `COPY . /airplane` be a subset of `PackageCopyCmds`?

leeweisberger (approved):

pkg/build/node.go
colinking: If the user doesn't have a `package.json`, one of the `PackageCopyCmds` creates a filler one.

yolken-airplane: Fixed.

pkg/examples/typescript/yarnworkspacespostinstall/pkg1/src/index.ts
yolken-airplane: Fixed.

pkg/build/node_test.go
yolken-airplane: Fixed.

pkg/build/node_externals.go
yolken-airplane: Done, and also fixed `hasWorkspaces` to match.

PR #328, opened 2022-10-05
Add Name() to export resource interface

calufornia:
## Description

There are parts of the local dev server that require the resource's name (we've been using the resource slug instead) - this adds a method to the export resource interface which returns the resource's name, similar to how `ID()` returns the resource's ID.

Also deletes some unused resource functions.

## Test plan
n/a


francescaleung (commented):
wait, this is dumb. Since everything already embeds BaseResource (& should), can we just have getters on the Resource interface & put those getters on the BaseResource?

i.e.
```
func (r BaseResource) GetName() string {
  return r.Name
}
```

& then you don't need to individually define it for everything else?

Sometimes I forget how struct composition works in Go.

calufornia:
> wait, this is dumb. Since everything already embeds BaseResource (& should), can we just have getters on the Resource interface & put those getters on the BaseResource?
> 
> i.e.
> 
> ```
> func (r BaseResource) GetName() string {
>   return r.Name
> }
> ```
> 
> & then you don't need to individually define it for everything else?
> 
> Sometimes I forget how struct composition works in Go.

Better yet, I think we can define a `BaseResource` interface, and then embed that interface in the `Resource` interface (not sure if that's what you meant by having the getters on the `Resource` interface), I'll update the PR

pkg/resources/resource.go
```diff
@@ -2,21 +2,40 @@ package resources
...
 
-	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
 )
 
+type IBaseResource interface {
+	GetKind() ResourceKind
+	GetID() string
+	GetSlug() string
+	GetName() string
```
francescaleung, line 13: can we deprecate Kind() & ID() on the Resource interface?

francescaleung (approved):
assuming the removed functions are unused, lgtm

pkg/resources/resource.go
```diff
@@ -36,56 +55,6 @@ type ResourceKind string
...
-	return GetAirplaneEnvFromFunc(ref, name, os.LookupEnv)
-}
-
-func GetAirplaneEnvFromFunc(ref string, name string, f EnvLookupFunc) string {
-	key := fmt.Sprintf("AIRPLANE_%s_%s", strings.ToUpper(ref), strings.ToUpper(name))
-	val, _ := f(key)
-	return val
-}
-
-func AirplaneResourceFromFunc(ref string, f EnvLookupFunc, res Resource) error {
```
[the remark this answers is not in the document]
calufornia: yeah! only `GetAirplaneResourceFromFunc` is used

pkg/resources/resource.go
calufornia: yep, we should change any callers of `Kind()` and `ID()` to use `GetKind()` and `GetID()` before we remove them though. Ideally the base resource interface methods would just be `Kind()`, `ID()`, etc. but before we can do that we'll need to make the Base Resource methods private (or change their names) so as not to have the field and function names conflict

calufornia: I'll mark them as deprecated.

PR #330, opened 2022-10-05
Sanitize workflow interceptor logs and make shim quoting consistent

yolken-airplane:
## Description
This change sanitizes workflow interceptor logs by replacing any values associated with the keys `["airplane_token", "airplane_resources", "password", "token"]` with `[OMITTED]` so that we don't leak secrets into logs. This change also makes the quoting style consistent across all of the shim files.

Resolves TSK-822

## Test plan
Testing completed successfully via building and running locally and verifying that logs were sanitized as expected:

Screen Shot 2022-10-05 at 4 26 31 PM


pkg/build/workflow/worker-shim.js
```diff
@@ -148,10 +150,18 @@ function activityLog(info, message) {
   );
 }
 
+function logStringify(obj) {
+  return JSON.stringify(obj, (key, value) =>
+    sensitiveLogKeys.includes(key.toLowerCase()) ? "[OMITTED]" : value
```
colinking, line 155: TIL you can do this. Neat!

colinking (approved):
We should formalize a notion of "secret values" that can be automatically omitted from logs -- let's chat about that, too.

pkg/build/workflow/worker-shim.js
yolken-airplane: Yeah, I didn't realize either until I found it in the docs. :-)

PR #334, opened 2022-10-07
Use workflow shims from SDK

yolken-airplane:
## Description
This change updates the workflow shims in lib to use the SDK-provided interceptors and worker code added in https://github.com/airplanedev/js-sdk/pull/159. Since the new shims only work with newer SDK versions, this change also adds a version check in the workflow case and fails the build with an error if the SDK version in the user's bundle isn't high enough.

**NOTE:** The error message is shown correctly when building with `--local`, but is hidden when running through the builder service. I'll create a separate change in airport to fix this.

cc: @calufornia 

## Test plan
Testing completed successfully via deploying and running some tasks with the updated SDK code locally. I also verified that workflow builds with old SDK versions fail as expected:

```
Benjamin-Yolkens-MacBook-Pro:workflow benjaminyolken$ aplocal deploy --local
Deploying 1 task:

js_sdk_workflow
Type: node
Root directory: ./../..
URL: https://app.airplane.so:5000/t/js_sdk_workflow

⣟Building...

Error: The airplane SDK version in your bundle (0.2.9) does not meet the minimum required one for workflow tasks (0.2.10). Please update the version of your airplane dependency.

```


pkg/build/workflow/worker-shim.js
```diff
@@ -1,160 +1,4 @@
...
-
-function logStringify(obj) {
-  return JSON.stringify(obj, (key, value) =>
-    sensitiveLogKeys.includes(key.toLowerCase()) ? "[OMITTED]" : value
-  );
-}
+import { runWorker } from "@airplane/workflow-runtime/internal";
 
 // This function is the worker's entrypoint.
 async function main() {
```
colinking, line 4: Do we need to the `JSON.parse` logic? I wouldn't expect parameters to get passed into the worker.

colinking (changes requested):
LGTM, but let's add that version check before merging this. Let's make sure the error is nice and clear about how the user just needs to upgrade their SDK version.

pkg/build/workflow/worker-shim.js
yolken-airplane: Yeah, they're actually used. Will remove them...

yolken-airplane:
> LGTM, but let's add that version check before merging this. Let's make sure the error is nice and clear about how the user just needs to upgrade their SDK version.

Yup, will add that after my other change has been merged and I've bumped up the SDK version there.

pkg/build/workflow/worker-shim.js
yolken-airplane: Done.

pkg/build/node.go
```diff
@@ -380,9 +384,25 @@ func GenShimPackageJSON(packageJSONs []string, isWorkflow bool) ([]byte, error)
 	// Always keep the versions of airplane and @airplane/workflow-runtime in sync, unless the task's dependencies
 	// explicitly include @airplane/workflow-runtime.
 	if isWorkflow {
-		if version, containsAirplane := deps["airplane"]; containsAirplane {
+		if apVersion, containsAirplane := deps["airplane"]; containsAirplane {
+			// Make sure that the version is new enough to support latest
+			// workflow shims.
+			meetsMin, err := meetsMinimumVersion(apVersion, minWorkflowSDKVersion)
```
colinking, line 390: Hmm, I'm now realizing both this and the `workflowRuntimePkg` logic below are incorrect. The version we get from a `package.json` will be a range. We want the exact version used, which we would need to get from the lockfile.

For example, I might have a task with `airplane: ^0.2.3`. If I ran `npm install` when `0.2.4` was the latest version, then the lockfile will specify `0.2.4`. If I run `npm ci` (like our builder does), it'll install `0.2.4` even though `0.2.10` is the latest acceptable version by semver range. That means (1) the `meetsMinimumVersion` won't work and (2) the shim will have a different version of the workflow runtime since the shim copies the range but not the lockfile.

The proper behavior would be to extract the `airplane` version from the lockfile (falling back on the `package.json` version if there is no lockfile... though that may be unnecessary, since I think `npm ci`/`yarn install --frozen-lockfile` fail if there is no lockfile) and use that.

colinking: Looks like you can get this version using the npm/yarn CLIs:

```
yarn list --pattern airplane --silent --json --no-progress | jq -r '.data.trees[0].name'
cat package-lock.json | jq -r '.packages["node_modules/airplane"].version'
```

colinking: Hmm, judging by your `meetsMinimumVersion`, you're ignoring the `^/~` entirely s.t. `^0.2.3` will still error. That works, though it could have false positives. Maybe that's acceptable? Your call. If we change this now, it'll also benefit the `workflowRuntimePkg` logic :)

colinking (approved):

pkg/build/node.go
yolken-airplane: Yeah, I was trying to take an approach that was overly conservative (but not 100% correct). 

I've attempted to fix this by getting the version via yarn or `package-lock.json`; if a version isn't found there, we just fall back to the version from the deps. Does this seem ok to you?

pkg/build/node.go
```diff
@@ -14,9 +14,13 @@ import (
 	"github.com/pkg/errors"
 )
 
-const DefaultNodeVersion = "18"
-const defaultSDKVersion = "0.2.0"
-const workflowRuntimePkg = "@airplane/workflow-runtime"
+const (
+	DefaultNodeVersion = "18"
+
+	defaultSDKVersion     = "0.2.10"
```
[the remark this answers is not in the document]
yolken-airplane: Done.

pkg/build/node.go
```diff
@@ -380,9 +384,25 @@ func GenShimPackageJSON(packageJSONs []string, isWorkflow bool) ([]byte, error)
...
+			// workflow shims.
+			meetsMin, err := meetsMinimumVersion(apVersion, minWorkflowSDKVersion)
+			if err != nil {
+				return nil, err
+			}
+
+			if !meetsMin {
+				return nil,
+					errors.Errorf(
+						"The airplane SDK version in your bundle (%s) does not meet the minimum required one for workflow tasks (%s). Please update the version of your airplane dependency.",
```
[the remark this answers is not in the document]
yolken-airplane: Fixed.

pkg/build/node_externals.go
```diff
@@ -375,3 +376,138 @@ func getWorkspaceDependencyPackages(pathPackageJSON string) (map[string]bool, er
...
+
+	packageLock := npmPackageLock{}
+	if err := json.Unmarshal(contents, &packageLock); err != nil {
+		return "", errors.Wrap(err, "parsing package-lock.json")
+	}
+
+	// The package name is prefixed by "node_modules"
+	packageObj, ok := packageLock.Packages[fmt.Sprintf("node_modules/%s", packageName)]
+	if !ok {
+		return "", errors.Errorf("No version found for package %s", packageName)
```
colinking, line 450: nit: lowercase, for wrapping

```suggestion
		return "", errors.Errorf("no version found for package %s", packageName)
```

pkg/build/node_externals.go
```diff
@@ -375,3 +376,138 @@ func getWorkspaceDependencyPackages(pathPackageJSON string) (map[string]bool, er
...
+// file).
+func getLockPackageVersion(
+	rootDir string,
+	packageName string,
+	fallbackVersion string,
+) string {
+	yarnVersion, err := getYarnLockPackageVersion(rootDir, packageName)
+	if err == nil {
+		log.Printf(
+			"Found %s version via yarn: %s\n",
```
colinking, line 467: I generally try to use `%q` since it is more clear within sentences and handles empty strings

```suggestion
			"Found %q version via yarn: %q\n",
```

pkg/build/node_externals.go
```diff
@@ -375,3 +376,138 @@ func getWorkspaceDependencyPackages(pathPackageJSON string) (map[string]bool, er
...
+			packageName,
+			yarnVersion,
+		)
+		return yarnVersion
+	}
+	log.Printf(
+		"Error getting %s version from yarn: %+v\n",
+		packageName,
+		err,
+	)
```
colinking, line 477: These should be debug logs if anything. We don't want a user to see them in their build logs and think they did something wrong.

colinking (approved):
This is fantastic. Thanks!!

pkg/build/node_externals.go
yolken-airplane: They're actually completely hidden from users due to the logic in https://github.com/airplanedev/airport/blob/main/pkg/builder/builder.go#L671.

yolken-airplane: Yeah, actually, that makes sense given the code- anything that we log out outside of buildkit will be hidden by default unless it has the magical build log prefix.

pkg/build/node_externals.go
yolken-airplane: Fixed. So we never want to use upper-case for the first letter of an error message? I was wondering what the correct standard was there. :-)

pkg/build/node_externals.go
yolken-airplane: Oh, didn't know about that. Fixed!

pkg/build/node_externals.go
colinking: We should lint for it, but yeah -- lowercase always and optionally add logic to uppercase the first char when printing error messages. This is how it works in our API!

PR #335, opened 2022-10-07
Use slim image for building a view

leeweisberger:
Switch views to using a slim node image. This should significantly improve build time as the slim image gets pulled in a lot faster than the fat node version.

Technically this is backwards incompatible. However I highly doubt that anyone (in the past or probably the future) is going to need more than what the slim image provides to a) install dependencies and b) build the view. In the future we could allow the user to override this, but again I don't **think** this is going to happen except in rare cases

I think this is probably what we should have done for node tasks from the start, but we can at least do this with views.

Thoughts? Questions?

I also considered using an alpine image, but it's not too much smaller than the slim image and although the alpine image should work for 99% of cases, the slim image should work for 99.5% :)


yolken-airplane (approved):

zhan-airplane (approved):
OOC how much faster is it?

colinking:
This may cause issues for install hooks that need something from the debian image. However, I don't have an example off the top of my head. Your call!

If this is to improve build times, it would only help with pulling the base image (we push these base images to a public cache so builds do not need to push the base layers -- speaking of which, make sure to push this image into the public cache!). To improve build times for pulling base images, could we instead mount a cache of the base images to the builder? That would avoid any risk of breaking changes.

colinking (changes requested):
(pushing from my review queue)

zhan-airplane:
@leeweisberger are you planning to merge this? My vote would be to go through with it haha (better to do it earlier before more people start using views), and maybe have a flag somewhere to use the non-slim base image if they need something funky from it.

leeweisberger:
> @leeweisberger are you planning to merge this? My vote would be to go through with it haha (better to do it earlier before more people start using views), and maybe have a flag somewhere to use the non-slim base image if they need something funky from it.

Can we chat in person first since you're going to be working on perf? I will merge it unless you're going to tackle the larger project of mounting base images on our builder pods which would make this less necessary

leeweisberger:
@zhan-airplane and I discussed. I'm going to close for now and let him take over.

PR #337, opened 2022-10-11
Standardize parameter conversion logic

colinking:
## Description

The editor calls `GetParameters` on the task definition. However, that function didn't include the same conversion logic that we have elsewhere for converting between the API and definition-file parameter formats. For example, it didn't set constraints nor the component field.

cc @francescaleung @kellymtrinh 

Resolves TSK-854

## Test plan

Confirmed that the editor shows the correct parameter forms.


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -1056,97 +1057,6 @@ func (d Definition_0_3) GetUpdateTaskRequest(ctx context.Context, client api.IAP
 	return req, nil
 }
 
-func (d Definition_0_3) addParametersToUpdateTaskRequest(ctx context.Context, req *api.UpdateTaskRequest) error {
-	req.Parameters = make([]api.Parameter, len(d.Parameters))
```
colinking, line 1060: All of this parameter conversion logic was just extracted into the other file. No major changes apart from minor variable renames!

kellymtrinh (approved):

PR #340, opened 2022-10-12
Add SupportedDefinitionExtensions

kellymtrinh:
## Description
`SupportedFileExtensions` is unused except for hot reloading, which isn't out yet. 

Realized for hot reloading we don't care about all file extensions, just ones that can possibly contain task or view definitions.
This means inline configs for tasks, yaml for tasks and views, and files accompany yaml definitions and get loaded in to the def (which should just be SQL?). 

We don't want to refresh the code JS/python/non inline view files since those already get updated either at execution time or by vite directly.

CC: @ameyakhare @leeweisberger 

## Test plan
- N/A, unused.


pkg/deploy/discover/discover.go
```diff
@@ -21,15 +21,18 @@ var IgnoredDirectories = map[string]bool{
...
-	".ts",
-	".tsx",
-	".js",
-	".jsx",
+// DefinitionFileExtensions is all the file extensions that could contain
+// a task or view definition: inline task or view configs, or files that accompany yaml definitions
+var DefinitionFileExtensions = []string{
+	// yaml configs and associated files
 	".yaml",
+	".sql",
```
calufornia, line 29: Maybe add a comment here on why we include `.sql` - since the SQL gets loaded into the task definition

calufornia (approved):

PR #344, opened 2022-10-13
Add template evaluation request and response

calufornia:
## Description

These will be used by the CLI and our API, so adding to lib. Also removes the interpolation logic from lib for SQL + REST tasks, since this will be part of the CLI.

cc @kellymtrinh 

## Test plan
n/a


pkg/api/client.go
```diff
@@ -519,3 +519,30 @@ type GetEnvResponse struct {
...
+	//     "my_param": "hello",
+	//     "my_other_param": "hi"
+	//   },
+	//   "configs": {
+	//     "FOO": "BAR"
+	//   }
+	// }
+	// Any references to params.my_param, params.my_other_param, or configs.FOO
+	// in Value will be replaced by the values above.
+	LookupMaps map[string]map[string]interface{} `json:"lookupMaps"`
```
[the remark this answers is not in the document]
calufornia: Just to confirm, do you mean something like:
```
type EvaluateTemplateRequest struct {
    Value interface{} `json:"value"`
    Run `json:"Run"`
    Env `json:"Env"`
    ...
```
Maybe we support both this and the generic lookup map?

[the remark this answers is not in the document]
calufornia: So to start, we'd have something like:
```
type EvaluateTemplateRequest struct {
    Value interface{} `json:"value"`
    Run
    Env
    Resources
    Params
    Configs
    LookupMaps map[string]map[string]interface{} `json:"lookupMaps"`
```

pkg/api/client.go
```diff
@@ -530,3 +531,77 @@ type GetEnvResponse struct {
...
 }
+
+type EvaluateTemplateRequest struct {
+	// Value is an arbitrary value that can include one or more Template values.
+	// Each Template will be evaluated, and if successful, will be replaced in
+	// Value with its output. The updated Value will be returned in the response.
+	// If any templates fail to evaluate, the Template will be left in Value
+	// and a separate error will be returned in the response.
+	Value       interface{}                   `json:"value"`
+	RunID       string                        `json:"runID"`
```
francescaleung, line 542: Maybe you want this to be a Run object instead? If it's a local run it won't be available from the database.

pkg/api/client.go
```diff
@@ -530,3 +531,77 @@ type GetEnvResponse struct {
...
+	Env         Env                           `json:"env"`
+	Resources   map[string]resources.Resource `json:"resources"`
+	Configs     map[string]string             `json:"configs"`
+	ParamValues map[string]interface{}        `json:"params"`
```
francescaleung, line 546:
```suggestion
	ParamValues map[string]interface{}        `json:"paramValues"`
```

Unless there's a really good reason for this to be `params`, best to be consistent here.

pkg/api/client.go
```diff
@@ -530,3 +531,77 @@ type GetEnvResponse struct {
...
+	//   "params": {
+	//     "my_param": "hello",
+	//     "my_other_param": "hi"
+	//   },
+	//   "configs": {
+	//     "FOO": "BAR"
+	//   }
+	// }
+	// Any references to params.my_param, params.my_other_param, or configs.FOO
+	// in Value will be replaced by the values above.
```
francescaleung, line 559: Can you edit this example & clarify that these are _in addition_ to lookup maps provided by the objects above (so `params` and `configs` would be covered by `ParamValues` & `Configs` respectively)? I assume they'll override the lookups if there are collisions, you should probably make note of that as well.

pkg/runtime/sql/sql.go
```diff
@@ -33,16 +33,12 @@ type Runtime struct{}
 
 // PrepareRun implementation.
 func (r Runtime) PrepareRun(ctx context.Context, logger logger.Logger, opts runtime.PrepareRunOptions) (rexprs []string, rcloser io.Closer, rerr error) {
-	kindOptions, err := runtime.EvalRunOptionTemplates(ctx, opts)
-	if err != nil {
-		return nil, nil, err
-	}
```
francescaleung, line 39: is this being removed because interpolation is being applied to all run inputs now?

francescaleung (approved):

pkg/runtime/sql/sql.go
kellymtrinh: Yep, this is the old eval logic that is going to be deleted

pkg/api/client.go
calufornia: We don't actually use the run ID to look up the run in the API - we just set `run.id` to be this value. But we should eventually add a `Run` struct to lib and use that here and in airport instead, I'll add a TODO.

PR #345, opened 2022-10-13
Implement bundle based discovery

leeweisberger:
## Description
Implement a new discoverer that discovers bundles rather than tasks/views.

This discoverer is similar to the existing one, but it doesn't go so far as to get view/task configs. For each type of discoverer, it gets the root of the potential task/view.

Each bundle stores 
1. The path to the root of the bundle.
2. The target paths relative to the root that match entities that should be deployed.
3. (TODO) the build type and version of the bundle.

## Test plan
I added unit tests with fixtures. Nobody is using this yet.


pkg/deploy/utils/entities.go
```diff
@@ -0,0 +1,33 @@
+package deployutils
```
leeweisberger, line 1: This ended up not really being used, but we're copying this all over the place we need a source of truth.

@kellymtrinh we should use this instead of the map you created.

kellymtrinh: Oh yes this would be great!!

ameyakhare (approved):

PR #346, opened 2022-10-14
Fix json struct tag when unmarshaling EvaluateTemplateRequest

calufornia:
## Description
This was changed to `"paramValues"` in the request, we fix the raw message's struct tag to be this as well.

## Test plan
Test with lib bump in API.


pkg/api/client.go
```diff
@@ -572,7 +572,7 @@ func (r *EvaluateTemplateRequest) UnmarshalJSON(buf []byte) error {
 		Env         Env                               `json:"env"`
 		Resources   map[string]map[string]interface{} `json:"resources"`
 		Configs     map[string]string                 `json:"configs"`
-		ParamValues map[string]interface{}            `json:"params"`
```
kellymtrinh, line 575: Do both need to be ``json:"params"`` instead? Because eval will look for `params.foo` not `paramValues.foo`?

calufornia: Ah so the struct tag has no bearing on the global name - that's set in the API (we just take the request's paramValues and create a `"params"` lookup map

kellymtrinh (approved):

PR #347, opened 2022-10-14
Delete expressions package

kellymtrinh:
## Description
This is unused now since we moved to using the eval templates endpoint: https://github.com/airplanedev/lib/pull/344

## Test plan
- Using CLI with this version of lib compiles.
- Using airport with this version of lib compiles.


calufornia:
Can you also delete this: https://github.com/airplanedev/lib/blob/86d76a07b805893559ba24ff01d50118beed0af4/pkg/runtime/runtime.go#L107

PR #349, opened 2022-10-14
Add TaskID to UpdateTaskRequest type

### LINKED ISSUE #348: TaskID missing from UpdateTask type
TaskID is missing from UpdateTask type.  This prevents updating the slug for a task as in the UI.

freimer:
## Description

This change adds TaskID to UpdateTaskRequest so the slug can be changed as in the UI.

## Test plan

I believe the UI uses /i/tasks/update whereas the api in airplanedev/cli/pkg/api/client.go uses /v0/tasks/update, so I'm actually not sure if this will work. The ultimate goal is to allow API clients to update the task slug for an existing task like the UI does. So this will need to be tested. Marking as draft until those tests are completed.

Closes #348


freimer:
This does not work with the /v0/tasks/update endpoint, so the api will actually need to be changed to support changing the task slug via the API as the UI can do.

PR #350, opened 2022-10-14
Do not set `nodeVersion` for workflows

colinking:
## Description

When initializing a workflow, don't set `nodeVersion`. Since workflows do not run in a Node runtime (they run in V8), the `nodeVersion` field has no impact.

For backwards compatibility, we still accept a `nodeVersion` field within inline workflow config. This field was already optional (even for tasks, which differs from the YAML format, but that's a separate discussion to have), so we don't need to change anything.

cc @kellymtrinh @ameyakhare 

Resolves TSK-684

## Test plan

Initialized a task and confirmed it had `nodeVersion` set. Initialized a workflow and confirmed it did not have `nodeVersion` set.


colinking:
* main
  * **PR #350** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/350" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/350?utm_source=stack-comment).

ericliuche (approved):

PR #351, opened 2022-10-14
Add build type and version to bundle

leeweisberger:
## Description
When discovering bundles, also add the build type and version to the bundle.

Bundles should be uniquely created based on their root path, build type, and build type version. i.e. two bundles with the same root might be created for different versions of a build type.

## Test plan
Unit tests with fixtures.


pkg/deploy/bundlediscover/discover.go
```diff
@@ -136,27 +167,12 @@ func updateBundleWithTarget(b Bundle, target string) (Bundle, error) {
...
-		}
-		isSubDirectoryOfExistingPath, err := fsx.IsSubDirectory(existingTargetPath, relPath)
-		if err != nil {
-			return b, err
-		}
-		if isSubDirectoryOfExistingPath {
-			shouldAdd = false
-			break
-		}
-	}
```
leeweisberger, line 157: Decided this wasn't necessary. A bundle may come with a target path that is a child of another, this is ok still valid

pkg/deploy/bundlediscover/discover.go
```diff
@@ -108,26 +108,57 @@ func (d *Discoverer) Discover(ctx context.Context, paths ...string) ([]Bundle, e
...
 	}
 
-	var bundles []Bundle
-	for _, b := range bundleByPath {
-		bundles = append(bundles, b)
+	return bundles, nil
+}
+
+func addBundle(bundles []Bundle, bundle Bundle, path string) ([]Bundle, error) {
+	matchingBundle := -1
```
ameyakhare, line 128: maybe just use a bundle pointer here instead of an index thing

ameyakhare (approved):

PR #352, opened 2022-10-17
Add created at and updated at fields to Tasks

kellymtrinh:
## Description
Adds these two fields to match the API `Task` struct https://github.com/airplanedev/airport/blob/main/pkg/ap/tasks.go#L47 - previously was not needed.

This is because we want to start showing "last updated at" in the studio.

## Test plan
- N/A, unused.


pkg/api/client.go
```diff
@@ -50,6 +50,10 @@ type Task struct {
 	IsArchived                 bool               `json:"isArchived" yaml:"isArchived"`
 	InterpolationMode          string             `json:"interpolationMode" yaml:"-"`
 	Triggers                   []Trigger          `json:"triggers" yaml:"-"`
+
+	CreatedAt time.Time `json:"createdAt"`
+	// Computed based on the task's revision.
+	UpdatedAt time.Time `json:"updatedAt"`
```
leeweisberger, line 56: Can you add `yaml:"-"` to be consistent with the other fields? I'm not 100% sure we need but maybe we do

leeweisberger (approved):

calufornia (approved):

PR #362, opened 2022-10-25
Add bundle build dockerfile generation

ameyakhare:
Resolves TSK-889

## Description
Adds a unique dockerfile for generating an image from a bundle build.

## Test plan
Works locally.


pkg/build/builder.go
```diff
@@ -333,6 +341,15 @@ func BuildDockerfile(c DockerfileConfig) (string, error) {
 	}
 }
 
+func BuildBundleDockerfile(c BundleDockerfileConfig) (string, error) {
+	switch c.BuildType {
+	case NodeBuildType:
```
leeweisberger, line 346: Going to add other types after?

pkg/build/node.go
```diff
@@ -621,3 +629,287 @@ func (p *pkgJSONWorkspaces) UnmarshalJSON(data []byte) error {
 	p.workspaces = workspacesObject.Packages
 	return nil
 }
+
+// node creates a dockerfile for Node (typescript/javascript).
```
leeweisberger, line 633: Use the correct function name. What's the diff between this and the node function?

pkg/build/node.go
```diff
@@ -621,3 +629,287 @@ func (p *pkgJSONWorkspaces) UnmarshalJSON(data []byte) error {
...
+
+		if err := json.Unmarshal(buf, &pkg); err != nil {
+			return "", fmt.Errorf("node: parsing %s - %w", rootPackageJSON, err)
+		}
+	}
+
+	for i, a := range buildArgs {
+		buildArgs[i] = fmt.Sprintf("ARG %s", a)
+	}
+	argsCommand := strings.Join(buildArgs, "\n")
```
leeweisberger, line 693: Split this into a helper and reuse

pkg/build/node.go
```diff
@@ -621,3 +629,287 @@ func (p *pkgJSONWorkspaces) UnmarshalJSON(data []byte) error {
...
+	if ok {
+		// Depending on how the options were serialized, the runtime can be
+		// either a string or TaskRuntime; handle both.
+		switch v := runtime.(type) {
+		case string:
+			isWorkflow = v == string(TaskRuntimeWorkflow)
+		case TaskRuntime:
+			isWorkflow = v == TaskRuntimeWorkflow
+		default:
+		}
```
leeweisberger, line 671: Split into a helper and reuse

pkg/build/node.go
```diff
@@ -621,3 +629,287 @@ func (p *pkgJSONWorkspaces) UnmarshalJSON(data []byte) error {
...
+		NodeVersion:         GetNodeVersion(options),
+		PreInstallCommand:   pkg.Settings.PreInstallCommand,
+		PostInstallCommand:  pkg.Settings.PostInstallCommand,
+		Args:                argsCommand,
+		IsWorkflow:          isWorkflow,
+		InlineParserScript:  parser.NodeParserScript,
+		InlineUniversalShim: inlineString(universalNodeShim),
+	}
+
+	// TODO: install hooks per root directory
```
leeweisberger, line 710: What does this mean? You're doing install hooks below. What are we missinig?

pkg/build/node.go
```diff
@@ -621,3 +629,287 @@ func (p *pkgJSONWorkspaces) UnmarshalJSON(data []byte) error {
...
+			// image lean. This doesn't apply to Yarn v2 (specifically Plug'n'Play), which uses the
+			// cache directory for storing packages.
+			installCommand = "yarn install --non-interactive --production --frozen-lockfile && yarn cache clean"
+		}
+	} else if hasPackageLock {
+		// Use npm ci if possible, since it's faster and behaves better:
+		// https://docs.npmjs.com/cli/v8/commands/npm-ci
+		installCommand = "npm ci --production"
+	}
+	cfg.InstallCommand = strings.ReplaceAll(installCommand, "\n", "\\n")
```
leeweisberger, line 812: break this into a helper

pkg/build/node.go
```diff
@@ -621,3 +629,287 @@ func (p *pkgJSONWorkspaces) UnmarshalJSON(data []byte) error {
...
+			&& {{.InlineWorkflowBundlerScript}} >> /airplane/.airplane/workflow-bundler.js
+		RUN node /airplane/.airplane/workflow-bundler.js
+		{{end}}
+
+		RUN {{.InlineUniversalShim}} > /airplane/.airplane/universal-shim.js && \
+			esbuild /airplane/.airplane/universal-shim.js \
+				--bundle \
+				--platform=node {{.ExternalFlags}} \
+				--target=node{{.NodeVersion}} \
+				--outfile=/airplane/.airplane/dist/universal-shim.js
```
leeweisberger, line 904: You should just build this once and then just copy it into the image. We don't need to build this for every single task that is deployed.

pkg/build/node.go
```diff
@@ -621,3 +629,287 @@ func (p *pkgJSONWorkspaces) UnmarshalJSON(data []byte) error {
...
+
+		RUN esbuild {{.BuildFilesList}} \
+			--bundle \
+			--platform=node {{.ExternalFlags}} \
+			--target=node{{.NodeVersion}} \
+			--outdir=/airplane/.airplane \
+			--outbase=/airplane
+
+		RUN node /airplane/.airplane/inlineParser.js {{.DiscoverFilesList}}
```
leeweisberger, line 913: How are we going to get this out of the image?

pkg/build/node.go
```diff
@@ -43,6 +44,10 @@ type templateParams struct {
 	PreInstallPath                   string
 	PostInstallPath                  string
 	PackageCopyCmds                  []string
+	BuildFilesList                   string
+	DiscoverFilesList                string
```
leeweisberger, line 48: Can you add comments what these are? There's a loooota stuffs here

pkg/build/node.go
leeweisberger: We can do as a followup, but there's no reason to do this every time. We can just add a build script next to the shim that produces the built js file.

leeweisberger (changes requested):

pkg/build/builder.go
ameyakhare: yep, thats the plan

pkg/build/node.go
ameyakhare: I fixed this. basically, we're only going to look for a pre/post path in the task root for bundle builds

pkg/build/node.go
ameyakhare: follow up, not doing it right now

leeweisberger (changes requested):
re-request after tests plz

pkg/build/node.go
```diff
@@ -44,6 +45,13 @@ type templateParams struct {
 	PreInstallPath                   string
 	PostInstallPath                  string
 	PackageCopyCmds                  []string
+
+	// string of space-separated entity files to esbuild (user code)
+	FilesToBuild string
+	// string of space-separated built js files to discover entity configs from
```
leeweisberger, line 51:
```suggestion
	// FilesToDiscover represents the built JS files to discover entity configs from. These built files are the output of esbuild.
```

pkg/build/node.go
```diff
@@ -654,3 +632,317 @@ func ReadPackageJSON(path string) (PackageJSON, error) {
...
+				--outfile=/airplane/.airplane/dist/universal-shim.js
+
+		RUN esbuild {{.FilesToBuild}} \
+			--bundle \
+			--platform=node {{.ExternalFlags}} \
+			--target=node{{.NodeVersion}} \
+			--outdir=/airplane/.airplane \
+			--outbase=/airplane
+
+		RUN node /airplane/.airplane-build-tools/inlineParser.js {{.FilesToDiscover}}
```
leeweisberger, line 890:
```suggestion
                # Discover inline tasks now that dependencies are installed and tasks are built.
                # FilesToDiscover should be the location where esbuild outputs transpiled JS files.
		RUN node /airplane/.airplane-build-tools/inlineParser.js {{.FilesToDiscover}}
```

leeweisberger (approved):

PR #366, opened 2022-10-27
Airplane config

leeweisberger:
## Description
Create airplane config yaml. 

Currently the airplane config only contains the node version and env vars.

The node version is read in when determining the node version (superseded by package.json).

So far nobody reads env vars. We are going to read these during dev and in the server during build. I'll add this functionality as follow ups.

cc @calufornia example of adding a jsonschema which we want to do for the airplane.dev.yaml

This is undocumented, so this should be a no-op.

## Test plan
Unit tests


pkg/deploy/config/config.go
```diff
@@ -0,0 +1,68 @@
...
+const (
+	FileName = "airplane.config.yaml"
+)
+
+//go:embed schema.json
+var schemaStr string
+
+type AirplaneConfig struct {
+	NodeVersion build.BuildTypeVersion `yaml:"nodeVersion,omitempty"`
+	EnvVars     api.TaskEnv            `yaml:"envVars,omitempty"`
```
ameyakhare, line 26: are these task env vars or build env vars? I feel like we should make the distinction that this is a buildenv var and task env vars should be passed in using the task config

ameyakhare (changes requested):
one open question, everything else looks fine

pkg/deploy/config/config.go
leeweisberger: I'm thinking they're both.
- We want to support build time env vars because those just won't work for inline tasks.
- It's also nice to support task env vars because this gives us a place to share env vars across many tasks. e.g. if all of our tasks uses `AIRPLANE_API` I can set it once here instead of on every individual task.

So we document that runtime env vars can optionally go here to share them amongst many tasks. Build time env vars have to go here.

Thoughts? too confusing? We could split them up, but I don't see a great reason to...

ameyakhare: I actually think we should split them up, since for environments, the idea is that those build env vars cannot be environment-specific

ameyakhare: not going to hard block on this, but its a little weird that build env vars and runtime env vars are lumped together when they really shouldnt need to be the same

ameyakhare (approved):

pkg/deploy/config/config.go
leeweisberger: I'm going to merge for now, but let's discuss. This isn't going to be shipped for a bit.

PR #371, opened 2022-10-27
Add builtins to `lib`

francescaleung:
## Description

Also adds GraphQL to the definition, which enables initializing + deploying a GraphQL task. The larger patch here is to accommodate builtin tasks generally.

Note that this does not generate a commented YAML file for builtins. Punting on this, because I think it might be harder than it's worth right now.

Resolves TSK-946.

## Test plan

Tested locally with updated CLI + API.


pkg/builtins/builtins.go
```diff
@@ -35,3 +48,39 @@ func GetBuiltinFunctionSpecification(slug string) (FunctionSpecification, error)
 		Name:      namespaceParts[1],
 	}, nil
 }
+
+// Returns a FunctionSpecification from a set of KindOptions. Expects a `functionSpecification` key
+// at the top level to contain a function specification serialized as a map[string]interface{}.
+func GetFunctionSpecificationFromKindOptions(kindOptions build.KindOptions) (FunctionSpecification, error) {
```
leeweisberger, line 54: Can you add a test for this plz

pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -233,7 +282,8 @@
     "shell": true,
     "docker": true,
     "sql": true,
-    "rest": true
+    "rest": true,
+	"graphql": true
```
leeweisberger, line 286: Is this weirdly indented or is that just in my github?

pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -213,6 +213,55 @@
...
+            "graphql": {
+              "description": "Configuration for a GraphQL task.",
+              "type": "object",
+              "properties": {
+                "resource": {
+                  "description": "The slug of a GraphQL resource.",
+                  "type": "string"
+                },
+                "query": {
+                  "description": "The set of GraphQL queries and mutations to use. This can contain multiple queries and mutations.",
```
leeweisberger, line 232: Maybe `this can contain multiple operations`? That seems to be the terminology used more commonly.

pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -213,6 +213,55 @@
...
+				  "type": "string"
+                },
+                "operationName": {
+                  "description": "The name of the operation to execute. This can be left blank if there is only a single operation in the query field.",
```
leeweisberger, line 236: `The name of the operation to execute from the query field`. Just to make clear that this is an operation name pulled from the query.

pkg/deploy/taskdir/definitions/builtins_graphql.go
```diff
@@ -0,0 +1,181 @@
...
+	"log"
+
+	"github.com/airplanedev/lib/pkg/api"
+	"github.com/airplanedev/lib/pkg/build"
+	"github.com/airplanedev/lib/pkg/builtins"
+	"github.com/pkg/errors"
+)
+
+func init() {
+	for _, plugin := range []TaskBuiltinPlugin{
```
leeweisberger, line 14: nit; it's a little hard to read. Can you pull `[]TaskBuiltinPlugin{` into a variable?

pkg/deploy/taskdir/definitions/builtins_graphql.go
```diff
@@ -0,0 +1,181 @@
...
+			func() BuiltinTaskDef { return &GraphQLDefinition_0_3{} },
+		),
+	} {
+		if err := registerBuiltinTaskPlugin(plugin); err != nil {
+			log.Fatal(err)
+		}
+	}
+}
+
+type GraphQLDefinition_0_3 struct {
```
leeweisberger, line 32: Should we name this `GraphQLRequestDefinition_0_3`?

pkg/deploy/taskdir/definitions/builtins_graphql.go
```diff
@@ -0,0 +1,181 @@
...
+
+func (d GraphQLDefinition_0_3) getAbsoluteEntrypoint() (string, error) {
+	return "", ErrNoEntrypoint
+}
+
+func (d GraphQLDefinition_0_3) getKindOptions() (build.KindOptions, error) {
+	variables := d.Variables
+	if variables == nil {
+		variables = map[string]interface{}{}
+	}
```
leeweisberger, line 136: Do we need to do this? Can it not be nil?

leeweisberger (approved):

pkg/deploy/taskdir/definitions/builtins_graphql.go
francescaleung: mehh, we're keying it to `graphql` in the definition, SQL/REST are both keyed to `sql` & `rest` respectively, this is probably fine. If we ever did MongoDB it would just be `MongoDBDefinition`.

pkg/deploy/taskdir/definitions/builtins_graphql.go
francescaleung: We have this for REST; I have vague memories of it being a thing. We can work around this--I think the material change here is that we get explicit null fields in the db, instead of the field being an empty object. I think empty object is semantically what we want here--i.e., there are no variables, so `variables` should be an empty map, not null. In practice we already handle these the same way on the frontend, so no reason why not to handle them the same way in other places...

My only reason for handling things this way is that there's already prior art for handling it this way, I don't know that I really have strong feelings about this.

PR #372, opened 2022-10-27
Fix shell task with newlines in params

leeweisberger:
The shell task shim fails to parse params if there are newlines in params.

This was reported by a user and I could repro from within the test.

For some reason the `cut` command behaves differently in the ubuntu machines we are running on and my machine. This all works fine on my machine but not on the ubuntu machine. After adding [-s](https://man7.org/linux/man-pages/man1/cut.1.html), it works correctly.

-f is documented as:
```
select only these fields;  also print any line that
              contains no delimiter character, unless the -s option is
              specified
```

I think the last part (also print any line) is the culprit. For something like `myParam=first\nsecond`, we were getting `myParam=first` as the first split. I'm not 10000% sure... but it now works and is unit tested.

Error from customer:
`
.airplane/shim.sh: line 7: export: `PARAM_BARCODES
XYZ333=ABC123
XYZ333': not a valid identifier
`

FIXES TSK-536

Once this is checked in, this run should run correctly: https://web.airstage.app/runs/run20221027zwsjq1iqdlr?__env=prod


pkg/build/builder_test.go
```diff
@@ -97,13 +97,15 @@ func RunTests(tt *testing.T, ctx context.Context, tests []Test) {
 func runTask(t *testing.T, ctx context.Context, dclient *client.Client, image string, paramValues Values) []byte {
 	require := require.New(t)
 
-	pv, err := json.Marshal(paramValues)
-	require.NoError(err)
+	var params []string
+	for k, v := range paramValues {
+		params = append(params, fmt.Sprintf("%s=%s", k, v))
+	}
```
leeweisberger, line 103: This was never working correctly. Tests were passing because they were producing an error that contained the parameter and we are searching for the parameter as a string in this test which was contained in the error!

pkg/examples/shell/ubuntu-no-newline/main.sh
```diff
@@ -4,4 +4,4 @@
 
 echo "Echoing env variable"
 
-echo "${foo}"
+env
```
leeweisberger, line 7: This test wasn't passing and got a newline added to it. It was passing because it caused an error that printed the param. Now it's actually printing the param.

pkg/build/builder_test.go
```diff
@@ -87,23 +88,33 @@ func RunTests(tt *testing.T, ctx context.Context, tests []Test) {
...
 
-	pv, err := json.Marshal(paramValues)
-	require.NoError(err)
+	var paramsStr string
+	if kind == TaskKindShell {
+		var params []string
+		for k, v := range paramValues {
+			params = append(params, fmt.Sprintf("%s=%s", k, v))
+		}
+		paramsStr = strings.Join(params, ", ")
```
leeweisberger, line 107: This was never working! shell tasks have params passed in differently than other tasks. Tests were passing because this produced an error which itself contained the target string we were looking for. My new tests would fail (correclty) without this

yolken-airplane (approved):

PR #375, opened 2022-10-28
Use smaller images for workflows

yolken-airplane:
## Description
This change updates the node builder to use slim base images for workflow builds. It also removes some large binaries from workflow images that aren't needed because they're for platforms we're not running on (e.g., windows). 

The new images are ~900MB smaller, and it seems like workflows run fine in them.

cc: @colinking @ericliuche 

## Test plan
Testing completed successfully via building a workflow locally with a slim base image and verifying that it ran correctly. Here's the before-and-after size for a workflow image:

```
us-central1-docker.pkg.dev/airplane-stage/team-1jq9ndetctmyqqxajesk7v0wxav/task-tsk20221007zkxi41sl2za   <none>                  8a7fba61db7d   31 hours ago         1.32GB
us-central1-docker.pkg.dev/airplane-stage/team-1jq9ndetctmyqqxajesk7v0wxav/task-tsk20221007zkxi41sl2za   latest                  de5a0d582bd0   About a minute ago   419MB
```


pkg/build/node.go
```diff
@@ -176,7 +176,12 @@ func node(
 		cfg.Workdir = "/" + cfg.Workdir
 	}
 
-	cfg.Base, err = getBaseNodeImage(cfg.NodeVersion)
+	// Always use slim images for workflows.
+	//
+	// TODO: Consider also using them for standard node workflows.
```
leeweisberger, line 181: Do you mean standard node tasks here?

pkg/build/versions.json
```diff
@@ -5,11 +5,21 @@
...
     "16": {
       "image": "registry.hub.docker.com/library/node",
       "tag": "16.16.0-bullseye",
       "digest": "sha256:e5cc51d25639e8ca6ee34ab8320680268a87103bafdcd48979adaa2e2815eae6"
     },
+    "16-slim": {
+      "image": "registry.hub.docker.com/library/node",
+      "tag": "16.16.0-bullseye-slim",
+      "digest": "sha256:cda7229eb72b7534396e7b58ba5b9f2454aee188317e058cbbf22686e5d07e2f"
+    },
```
leeweisberger, line 22: Any reason you didn't add 14? It is still a supported version

leeweisberger (approved):

pkg/build/versions.json
yolken-airplane: I was thinking that all workflows users would be on newer node versions. But, added slim images for the old ones (except 12) as well. I've also added a slim python-3 image for completeness even though there's no way of using it at the moment.

PR #376, opened 2022-10-28
Psql don't require password

leeweisberger:
## Description

psql should not require a password.

## Testing

I cannot figure out how to test this locally right now :) I can either get help or we can deploy this, apply it to airport, and try it out in stage.


joshma (approved):
Let's test it on stage and rollback ASAP if (for some reason) it doesn't work.

leeweisberger:
> Let's test it on stage and rollback ASAP if (for some reason) it doesn't work.

This shouldn't cause issues, just might run into another error downstream with no password

PR #377, opened 2022-10-31
Add option to skip verifying missing tasks in discoverers

ameyakhare:
## Description
We don't want to verify missing tasks/views/create them in the new bundler deployments flow, so we provide an option here to skip calling the api to ensure existence of the task/view.

## Test plan
👀


pkg/deploy/discover/code_view_discover.go
```diff
@@ -22,6 +22,10 @@ type CodeViewDiscoverer struct {
 	Logger logger.Logger
 
 	MissingViewHandler func(context.Context, definitions.ViewDefinition) (*api.View, error)
+
+	// DoNotVerifyMissingTasks will return ViewConfigs for views without verifying their existence
+	// in the api. If this value is set to true, MissingViewHandler is ignored.
+	DoNotVerifyMissingTasks bool
```
leeweisberger, line 28: DoNotVerifyMissingViews

leeweisberger (approved):

PR #379, opened 2022-11-01
Make choice of slim vs. full base images explicit

yolken-airplane:
## Description
This change updates our node and python task kind configs to add a `base` parameter that can be set to `"slim"` to indicate that we should use a slim base image when building. This change replaces the forced-slim behavior for workflows that was added in https://github.com/airplanedev/lib/pull/375.

The benefits of this change are that (1) non-workflow users can also use slim images and (2) workflow users can use non-slim images if they need to for whatever they're doing.

A few notes:

- I'm using a string instead of a bool so that we can possibly support other base image variants in the future (e.g., alpine vs. ubuntu)
- If base is unset, we use a non-slim image. Thus, the change should be a no-op for existing task configs. 
- Let me know if you think the term `base` isn't descriptive enough; we could use something like `baseImageType` instead?

cc: @colinking 

## Test plan
Testing completed successfully via extra unit tests and building locally with full and slim base configs.


pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -16,6 +16,7 @@ type TaskParam = {
 
 type NodeDef = {
   nodeVersion?: "12" | "14" | "15" | "16" | "18";
+  base?: "full" | "slim";
```
leeweisberger, line 19: Don't bother adding this. We're going to deprecate this anyways pretty much immediately. We're going to end up adding this to the `airplane.yaml` file that I'm adding. I'll handle that

leeweisberger: We're probably going to deprecate the option in the def file too, but that should continue to work even while deprecated while this will not work

pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -44,7 +49,12 @@
                   "description": "The path to the .py file containing the logic for this task. This can be absolute or relative to the location of the definition file.",
                   "type": "string"
                 },
-                "envVars": { "$ref": "#/$defs/envVars" }
+                "envVars": { "$ref": "#/$defs/envVars" },
+                "base": {
+                  "description": "The type of base image to use",
```
leeweisberger, line 54: Should we add that if not supplied, it will use the `full` image?

leeweisberger (approved):
What's our release strategy here? This let's people set the base as slim, but realistically nobody is going to know/opt into this unless we do it for them automatically.

pkg/deploy/discover/parser/node/parser.ts
yolken-airplane: So the benefit of doing this is that we can add it as an option into the js-sdk and then make it the default for workflow tasks. Is that a bad idea?

pkg/deploy/taskdir/definitions/schema_0_3.json
yolken-airplane: Fixed.

yolken-airplane:
> What's our release strategy here? This let's people set the base as slim, but realistically nobody is going to know/opt into this unless we do it for them automatically.

Good question. So, my thinking is that we roll it out in the following order:

1. Update the js-sdk to support specifying the base
2. Update our team's node tasks in airport to use slim base images; verify that nothing bad happens
3. Update the js-sdk to opt people into slim images by default (or maybe just workflow tasks?)
4. Update our node examples and templates that use a yaml task configs to set `base: slim` with a comment on what it means
5. Repeat the process for python tasks

Does that sound ok?

yolken-airplane:
> We should keep the happy path for task config to a minimum. Can we avoid needing to specify `base: "slim"`? What if we add that field and then do a breaking change (`0.3.0`) of the JS SDK where we make `slim` the default?

Sure, we can do that.

pkg/deploy/discover/parser/node/parser.ts
yolken-airplane: I don't mind if it's deprecated soon- I just want this in place so we can do some testing in the next few weeks. :-)

leeweisberger: This is fine, but adding this to the js sdk means that we have to add a breaking change (removing this config option) which isn't a huge deal since not many people are using inline tasks, but would rather avoid this if possible

leeweisberger:
> > We should keep the happy path for task config to a minimum. Can we avoid needing to specify `base: "slim"`? What if we add that field and then do a breaking change (`0.3.0`) of the JS SDK where we make `slim` the default?
> 
> Sure, we can do that.

We can do this as we roll out inline tasks? Default to slim for inline tasks so that when someone is migrating over they change to slim automatically. This is a good point to switch over!

pkg/deploy/discover/parser/node/parser.ts
yolken-airplane: Got it. I'll leave it undocumented and also not change any publicly-facing defaults or templates until the new deploy configuration is ready.

yolken-airplane: I can also add a comment in the js-sdk that this is an experimental feature that may break in the future.

yolken-airplane: Hmm, maybe I should just remove it for now then. This means we won't be able to test with workflows, though, since those are configured in code.

PR #386, opened 2022-11-02
Implement view bundle build

leeweisberger:
## Description
Add ability to build a bundle of views. This is similar but slightly different than tasks.

We're building the bundle as a [multi page app](https://vitejs.dev/guide/build.html#multi-page-app). The output of building will be the following file structure for a bundle with 3 views:

assets
- all js/css assets. i.e. the code
path/to/my/view/myview.view
- index.html -> runs main.tsx
- main.tsx -> imports from the assets folder
path/to/my/view/another.view
- index.html -> runs main.tsx
- main.tsx -> imports from the assets folder
anotherpath/aview
- index.html -> runs main.tsx
- main.tsx -> imports from the assets folder

We will upload this as one big application. If you were to serve it as one application, you would have to visit `https://www/path/to/my/view/myview.view/index.html` to serve the first view. We will automatically "route" the user to this path when serving them the view. This is just like how the agent automatically routes to the correct task when using the universal shim.

## Test plan
Unit tests confirm that a) this runs b) it produces the correct file path (by inspecting logs).
Have not yet e2e tested.


pkg/build/view.go
```diff
@@ -150,16 +153,234 @@ func view(root string, options KindOptions) (string, error) {
...
+		COPY package*.json yarn.* /airplane/
+		RUN {{.InlinePackageJSON}} > /airplane/package.json
+		RUN {{.InstallCommand}}
+
+		# Copy all source code to /src.
+		COPY . src/
+
+		{{if .FilesToDiscover}}
+		# Build and discover inline views.
+		RUN mkdir -p {{.DirectoryToBuildTo}}
```
ameyakhare, line 329: delet this

ameyakhare (approved):

PR #390, opened 2022-11-03
Support build base in airplane.yaml and bundle build

leeweisberger:
## Description
Note: both of these features are still in development....

1. Add `base` field to airplane.yaml (renamed in this pr from airplane.config.yaml per @joshma request). The user can specify a base in their `airplane.yaml` like they can set their build version. This value is overriden by any value they include in their defn file.

2. Add `BuildBase` to the bundle returned from bundle discovery. If there are two tasks with different bases, they will not be included in the same bundle. 

There will be a follow up in CLI and Airplort that uses the `BuildBase` returned in 2 and sends it to the API when building a bundle.

## Test plan
Unit tests updated.


pkg/build/node.go
```diff
@@ -178,7 +178,7 @@ func node(
 	}
 
 	baseImageType, _ := options["base"].(string)
-	cfg.UseSlimImage = baseImageType == "slim"
+	cfg.UseSlimImage = baseImageType == string(BuildBaseSlim)
```
leeweisberger, line 181: Switched to using an enum

pkg/deploy/bundlediscover/discover.go
```diff
@@ -127,7 +130,11 @@ func (d *Discoverer) Discover(ctx context.Context, paths ...string) ([]Bundle, e
 func addBundle(bundles []Bundle, bundle Bundle, path string) ([]Bundle, error) {
 	matchingBundle := -1
 	for i, b := range bundles {
-		if b.RootPath == bundle.RootPath && b.BuildType == bundle.BuildType && b.BuildVersion == bundle.BuildVersion {
+		match := b.RootPath == bundle.RootPath &&
```
leeweisberger, line 133: We now consider buildbase when determining if two bundles can be built together

pkg/deploy/config/config.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	FileName = "airplane.config.yaml"
+	FileName = "airplane.yaml"
```
leeweisberger, line 18: Changed name of `airplane.config.yaml` -> `airplane.yaml` (this isn't launched yet)

pkg/deploy/config/config.go
```diff
@@ -24,6 +24,7 @@ var schemaStr string
 type AirplaneConfig struct {
 	NodeVersion build.BuildTypeVersion `yaml:"nodeVersion,omitempty"`
 	EnvVars     api.TaskEnv            `yaml:"envVars,omitempty"`
+	Base        build.BuildBase        `yaml:"buildBase,omitempty"`
```
leeweisberger, line 27: Add Base to the airplane config yaml

pkg/deploy/discover/code_task_discover.go
```diff
@@ -103,9 +103,9 @@ func (c *CodeTaskDiscoverer) GetTaskConfigs(ctx context.Context, file string) ([
 	return taskConfigs, nil
 }
 
-func (c *CodeTaskDiscoverer) GetTaskRoot(ctx context.Context, file string) (string, build.BuildType, build.BuildTypeVersion, error) {
+func (c *CodeTaskDiscoverer) GetTaskRoot(ctx context.Context, file string) (string, build.BuildType, build.BuildTypeVersion, build.BuildBase, error) {
```
leeweisberger, line 106: All of these `GetXRoot` functions now also return the buildbase, taken from either the defn file or the airplane.yaml

pkg/deploy/discover/defn_discover.go
```diff
@@ -174,51 +174,56 @@ func (dd *DefnDiscoverer) GetTaskRoot(ctx context.Context, file string) (string,
...
 	}
-	if err := def.SetBuildVersion(taskPathMetadata.BuildVersion); err != nil {
-		return "", "", "", err
+	if err := def.SetBuildVersionBase(taskPathMetadata.BuildVersion, taskPathMetadata.BuildBase); err != nil {
+		return "", "", "", "", err
+	}
+	// Recalculate the build types.
+	buildType, buildTypeVersion, buildBase, err = def.GetBuildType()
+	if err != nil {
+		return "", "", "", "", err
```
leeweisberger, line 223: This was a bug that I fixed here. We need to recalculate these after setting a new version/base on the line above

pkg/deploy/discover/view_defn_discover.go
```diff
@@ -31,32 +31,12 @@ func (dd *ViewDefnDiscoverer) GetViewConfig(ctx context.Context, file string) (*
 		return nil, nil
 	}
 
-	buf, err := os.ReadFile(file)
+	d, err := getViewDefinitionFromFile(file)
```
leeweisberger, line 34: Refactored this into a helper so it can be reused

pkg/utils/pathcase/pathcase.go
```diff
@@ -26,6 +26,9 @@ func ActualFilename(filename string) (string, error) {
 		}
 	}
 
-	// If we haven't found the file, return the error (if any) from reading the directory.
-	return "", errors.Wrap(readErr, "listing directory")
+	if readErr != nil {
+		// If we haven't found the file, return the error (if any) from reading the directory.
+		return "", errors.Wrap(readErr, "listing directory")
+	}
+	return "", errors.Errorf("calculating actual filename for %s", filename)
```
leeweisberger, line 33: This was a bug I discovered due to a failing test. If readErr is nil, errors.wrap(readErr, "") returns nil.

pkg/build/node.go
```diff
@@ -783,8 +783,7 @@ func nodeBundle(
 		cfg.Workdir = "/" + cfg.Workdir
 	}
 
-	baseImageType, _ := options["base"].(string)
-	cfg.UseSlimImage = baseImageType == "slim"
+	cfg.UseSlimImage = buildContext.Base == BuildBaseSlim
```
leeweisberger, line 786: In the new bundle builds, we're going to deprecate this options thing and pull this off of the buildcontext. Again, this isn't launched yet and is behind a fflag

pkg/deploy/config/config.go
yolken-airplane: Should the YAML name be `base` instead of `buildBase`? Alternatively, we can change the other, user-facing instances of this to `buildBase` if you think that's cleaner. Just want to make sure we're consistent here.

pkg/build/node.go
yolken-airplane: Might be cleaner to just cast `options["base"]` directly to the `BuildBase` type?

yolken-airplane: Ditto for the python equivalent.

yolken-airplane (approved):
A few minor comments, otherwise good to merge.

pkg/deploy/config/config.go
leeweisberger: Yeah good catch. Base for now, we can keep thinking about it before we really launch this

PR #392, opened 2022-11-04
Implement bundle shell build

leeweisberger:
## Description

Implement shell bundle build.

To support `COPY --chmod` I switched us to using buildkit for builds (which we are using in the backend anyways so it's good to have parity).

## Test plan
Same unit tests as regular shell passes + a bunch of shell scripts bundled together.


pkg/build/bundle_builder.go
```diff
@@ -173,13 +175,33 @@ func (b *BundleBuilder) Build(ctx context.Context, bundleBuildID, version string
 
 	scanner := bufiox.NewScanner(resp.Body)
 	for scanner.Scan() {
-		var event *dockerJSONMessage.JSONMessage
-		if err := json.Unmarshal(scanner.Bytes(), &event); err != nil {
+		var msg *dockerJSONMessage.JSONMessage
```
leeweisberger, line 178: There is a better way to do this log displaying most likely. I spent most of my day trying to figure this out. This version is good enough imo for something that is mostly used in tests

pkg/build/shell.go
```diff
@@ -91,6 +91,73 @@ func shell(root string, options KindOptions) (string, error) {
...
+		`)
+		workDir = "/airplane"
+	}
+
+	// Extend template with our own logic - set up a WORKDIR and shim.
+	dockerfileTemplate = dockerfileTemplate + heredoc.Doc(`
+		WORKDIR {{.Workdir}}
+		RUN mkdir -p .airplane && {{.InlineShim}} > .airplane/shim.sh
+
+		COPY --chmod=700 . .
```
leeweisberger, line 149: Nice and simple!

pkg/build/universal-node-shim.js
```diff
@@ -3,19 +3,27 @@ import airplane from "airplane";
 
 async function main() {
   if (process.argv.length !== 5) {
-    console.log(`airplane_output_set:error ${JSON.stringify(`Expected to receive entrypoint, entrypointFunc, and params (via {{ "{{JSON}}" }}). Task CLI arguments may be misconfigured.`)}`);
+    console.log(
```
leeweisberger, line 6: I just formatted this file

pkg/build/universal-node-shim.js
```diff
@@ -3,19 +3,27 @@ import airplane from "airplane";
...
+      `airplane_output_set:error ${JSON.stringify(
+        `Expected to receive entrypoint, entrypointFunc, and params (via {{ "{{JSON}}" }}). Task CLI arguments may be misconfigured.`
+      )}`
+    );
     process.exit(1);
   }
 
+  const entrypoint = process.argv[2];
+  const entrypointFunc = process.argv[3];
+  const params = process.argv[4];
```
leeweisberger, line 16: And made these variables

pkg/build/builder_test.go
```diff
@@ -169,7 +181,9 @@ func runTask(t *testing.T, ctx context.Context, dclient *client.Client, c runTas
 	if len(c.Entrypoint) == 0 {
 		cmd = strslice.StrSlice{paramsStr}
 	} else {
-		c.Entrypoint = append(c.Entrypoint, paramsStr)
+		for k, v := range c.ParamValues {
+			c.Entrypoint = append(c.Entrypoint, fmt.Sprintf("%s=%s", k, v))
```
ameyakhare, line 185: umm I only think this works for bash, so you need to make this work for both

pkg/build/builder_test.go
```diff
@@ -204,6 +218,8 @@ func runTask(t *testing.T, ctx context.Context, dclient *client.Client, c runTas
 	logs, err := io.ReadAll(dlog.NewReader(logr, dlog.Options{
 		AppendNewline: true,
 	}))
+	fmt.Println("Output logs")
+	fmt.Println(string(logs))
```
ameyakhare, line 222: remove

ameyakhare (changes requested):

ameyakhare (approved):

PR #393, opened 2022-11-04
Remove operation name from graphql.request

francescaleung:
## Description

Also renames `Query` -> `Operation`, but only for the client.

Resolves TSK-1022.

## Test plan

Tested locally by initializing + deploying a GraphQL task.


pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -243,11 +243,7 @@
                   "type": "string"
                 },
                 "query": {
-                  "description": "The set of GraphQL queries and mutations to use. This can contain multiple operations.",
-                  "type": "string"
-                },
-                "operationName": {
-                  "description": "The name of the operation to execute from the query field. This can be left blank if there is only a single operation in the query field.",
+                  "description": "The query or mutation to execute.",
```
leeweisberger, line 246: Same here. We changing `query` -> `operation`?

I guess this would be a breaking change and require a backend change first...

leeweisberger (approved):

pkg/deploy/taskdir/definitions/schema_0_3.json
francescaleung: It would indeed be a breaking change. As far as I can tell no one has built a GraphQL task yet, so technically we can do this still, but even so I'm a little leery of doing it. Thoughts?

francescaleung:
@leeweisberger rerequesting for [cc709b4](https://github.com/airplanedev/lib/pull/393/commits/cc709b47674df60aabff267598e11928633895b5)

PR #397, opened 2022-11-07
Don't error if task-specific subdirectory already exists

calufornia:
## Description
We'll occasionally see local task execution flakiness due to the `.airplane` directory not getting cleaned up directory, which also leaves behind the task-specific subdirectory in `./airplane/{task_slug}`. In the case that this subdir exists, we ignore the error when attempting to create it.

This also fixes task execution in the general case where an `.airplane/` directory failed to get cleaned up (although this directory should get destroyed/recreated on subsequent studio startups anyway).

## Test plan
Tested in conjunction with studio.


leeweisberger (approved):

calufornia:
Looks like these changes were actually included in https://github.com/airplanedev/lib/commit/a322da11610ea83659e5f3c8e88acad61e0b0f59#diff-e56f9bb947923c6f8bb4c995bf4b9469a0ed117276223d16b177dcb04399a4bc - I'll still land this PR to update the comment on the edge case we were encountering.

PR #399, opened 2022-11-08
Support building python bundles

leeweisberger:
## Description
Support building python bundles using a universal shim.

## Test plan
Unit tests to ensure that we didn't break existing functionality and that new bundle based functionality is supported.


pkg/build/universal-node-shim.js
```diff
@@ -15,7 +15,6 @@ async function main() {
   const entrypointFunc = process.argv[3];
   const params = process.argv[4];
 
-  // TODO: needs environment variables to be set properly
```
ameyakhare, line 18: oops

pkg/build/python_test.go
```diff
@@ -69,3 +69,152 @@ func TestPythonBuilder(t *testing.T) {
...
+			ParamValues: map[string]interface{}{
+				"hello": "world",
+			},
+			BuildContext: BuildContext{
+				Type:    PythonBuildType,
+				Version: BuildTypeVersionPython37,
+			},
+			BundleRuns: []BundleTestRun{
+				{
+					RelEntrypoint: "main.py",
```
ameyakhare, line 93: why is there no export name for noninline tasks?

ameyakhare (approved):
looks fine

pkg/build/python_test.go
leeweisberger: This still works (both python and node) if entrypointFunc is not defined

PR #402, opened 2022-11-09
Add configs to top level of task definition

francescaleung:
## Description

Configs from SQL/REST definitions are only used if there are no configs from the top-level definition.

Resolves TSK-178.

## Test plan

Tested deploying + initializing REST tasks with configs attached, non-builtin tasks with env vars attached.


pkg/deploy/taskdir/definitions/schema_0_3.json
```diff
@@ -153,7 +153,10 @@
                   "default": "auto",
                   "enum": ["auto", "readOnly", "readWrite", "none"]
                 },
-                "configs": { "$ref": "#/$defs/configs" }
+                "configs": {
+                  "$ref": "#/$defs/configs",
+                  "description": "(Deprecated) A list of config variables that this task can access. Use top-level `configs` instead."
```
leeweisberger, line 158: There is a deprecated field in json schema.

Not 100% sure what it does, but we might as well use: http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.3

leeweisberger (approved):
Nice!

pkg/deploy/taskdir/definitions/schema_0_3.json
francescaleung: we're using an older draft, I think deprecated isn't available in draft 7.

PR #403, opened 2022-11-09
Install builtins at top level .airplane dir and update cleanup behavior

kellymtrinh:
## Description
Fixes issue where we run into a race condition where a task calling builtins might finish and cleanup the entire `.airplane` directory. This change puts the builtins into `.airplane/builtin-file` instead of `.airplane/builtins/builtin-file`. 
It will also only cleanup individual `.airplane/subtaskDir` instead of the entire `.airplane` directory, leaving the `.airplane/builtin-file`.

As discussed we'll revisit the cleanup behavior at a later time to only cleanup once the dev server closes. We don't want to cleanup the top level .airplane dir in any case, since it assumes that tasks always run sequentially and the first one to run is always the parent, which isn't true anymore.

This doesn't address possible race conditions between tasks.

## Test plan
I was able to reproduce the original issue by making a view that called a sql task and JS task with the same root:

Screen Shot 2022-11-09 at 3 30 07 PM
Confirmed that after this change it doesn't happen.

No harm:
cleared out all my .airplane dirs
- [x] test JS task executes
- [x] test view calling local rest task
- [x] task calling builtins
- [x] top level rest and sql task
- [x] task calling top level rest and sql tasks


pkg/utils/airplane_directory/airplane_directory.go
```diff
@@ -20,29 +20,34 @@ func (c closer) Close() error {
...
-	taskDir = filepath.Join(airplaneDir, slug)
-	if err := os.Mkdir(taskDir, os.ModeDir|0777); err != nil {
-		if !os.IsExist(err) {
-			return "", "", closer, errors.Wrap(err, "creating task subdirectory")
-		}
+	taskDir := filepath.Join(airplaneDir, slug)
+	if err := os.Mkdir(taskDir, os.ModeDir|0777); err == nil {
+		// Only set a closer if the current run caused the creation of the .airplane/task directory; else,
+		// defer removal of .airplane/task to the parent run.
+		closer = CloseFunc(func() error {
```
calufornia, line 46: Could you add a TODO here to remove this in favor of the dev server cleaning this up?

pkg/runtime/runtime.go
```diff
@@ -104,9 +104,8 @@ type PrepareRunOptions struct {
 	//
 	// It should be an absolute path.
 	Path string
-	// RootDir is the working directory the local run is being executed in
-	RootDir string
-
+	// WorkingDir is the working directory the local run is being executed in
+	WorkingDir string
```
calufornia, line 108: Is the task not executed from the root directory? i.e. if I run airplane dev in a nested folder without a package.json, doesn't it execute from one level up (where a package.json does exist)?

calufornia (approved):

pkg/runtime/runtime.go
kellymtrinh: I renamed to working directory instead of root dir because it's more relevant for the working directory used in studio, not necessarily the root dir of each task

PR #407, opened 2022-11-10
Tweak bundle discovery logic

leeweisberger:
## Description
I'm trying to solve two problems here:
1. Non built tasks (docker, rest, sql) always have a root of the tasks immediate parent directory.
2. Bundles that are children of other bundles will get uploaded separately even though the parent includes all of the code necessary for the children.

Both problems are outlined in this structure:
```
folder -> root for task1.js. Upload 1
    package.json
    airplane.yaml
    task1.js
    folder -> root for task2.sql. Upload 2
        task2.sql
        folder -> root for task3.sql. Upload 3
              task3.sql
    folder2 -> root for task4.js. Upload 4
        package.json
        airplane.yaml
        task4.js
```
Will produce 4 bundles all with different roots and different uploads. This means we create 4 code uploads. The root code upload (node) has the code for all of the tasks, but we will upload the other code bundles anyways. This is going to mess up the studio (cc @calufornia) and produce extra bundler pods and its generally confusing why task2 and task3 are built separately.

I have made two changes:
1. Non built tasks (docker, rest, sql) will "roll up" to the nearest `airplane.yaml` file if it exists. If not they'll follow current behavior. This ensures that task2.sql and task3.sql above get placed in the same bundle as each other and as the node root they're nested under. I think this intuitively makes sense as it "includes" these in the bundle that they are placed in.
2. Bundles that are children of another bundle should be uploaded together. They are still separate bundles. We are merging their roots and rewriting their target paths so they can share the same code but still build separately.

The above diagram is now:
```
folder -> root for task1.js, task2.sql, task3.sql. Upload 1
    package.json
    airplane.yaml
    task1.js
    folder
        task2.sql
        folder
              task3.sql
    folder2 -> root for task4.js. Upload 1
        package.json
        airplane.yaml
        task4.js
```
There are now 3 bundles instead of 4 (2 node bundles and 1 sql bundle). Moreover all of the bundles are uploaded with the same code, so there is only one code upload.


pkg/deploy/bundlediscover/discover.go
```diff
@@ -44,6 +44,64 @@ type Bundle struct {
...
+		if !alreadyAdded {
+			dedupedBundles = append(dedupedBundles, b)
+		}
+	}
+
+	// If any of the bundles are children of other bundles, convert the child root to being == the parent root.
+	// This ensures that we are creating as few bundles as possible.
+	// In order to do this, we set the child root path to the parent root path and update the target paths to be relative
+	// to the new (higher up) root.
+	for _, b := range dedupedBundles {
```
ameyakhare, line 80: im pretty sure you can only do this if the build type/version is the same?

ameyakhare: or will we be enforcing all of the subdirs in a bundle also have to have the same type/versions

ameyakhare (changes requested):
one open question, otherwise this looks fine

pkg/deploy/bundlediscover/discover.go
leeweisberger: This should actually be fine. The root here is the root that we are going to upload. If there's a python task and a node task and a sql task with the same root, they share the same code so they can be uploaded together. We still create a node/sql/python bundle (separate), each of which has the same root so should have the same upload.

I added another test for this and I think the behavior makes sense.,

leeweisberger: I'm basically trying to get bundles to have the same root so that they can share code. They're built separately from the same codebase based on their target paths & build type/version

ameyakhare (approved):

PR #408, opened 2022-11-10
Allow all python callables to be task candidates

ericliuche:
## Description
Allows for task to be defined as a variable, e.g.

```
import airplane

class MyClass:
    def __init__(self, foo: int, bar: int) -> None:
        self.foo = foo
        self.bar = bar

    def run(self, baz: int) -> None:
        print(self.foo, self.bar, baz)

# Define a task via a variable assignment
task_name = airplane.task()(MyClass(1, 2).run)
```

## Test plan
ran locally


pkg/deploy/discover/parser/python/parser.py
```diff
@@ -76,84 +75,83 @@ def as_def(obj: Any) -> Union[Any, Dict[str, Any]]:
...
-            tree = ast.parse(f.read(), filename=file)
-
         sys.path.append(os.path.dirname(file))
         spec = importlib.util.spec_from_file_location("airplane_task_import", file)
         assert spec is not None, f"Unable to import module {file}"
         assert spec.loader is not None, f"Unable to construct loader for module {file}"
         module = importlib.util.module_from_spec(spec)
         spec.loader.exec_module(module)
-        for item in tree.body:
-            if isinstance(item, ast.FunctionDef) and hasattr(module, item.name):
```
ericliuche, line 92: @ameyakhare what was your original reasoning for using `ast`?

ameyakhare: wasnt a deliberate decision

ameyakhare (approved):
whats the motivation for this? seems fine otherwise

ericliuche:
@ameyakhare panther asked for this https://airplanedev.slack.com/archives/C02KL1Y5QA2/p1667937569605239

PR #409, opened 2022-11-14
Support old tls version

leeweisberger:
## Description
When we upgraded to go1.18, we had to set a flag to support old TLS versions: https://github.com/airplanedev/airport/pull/2208

This flag is no longer supported in go1.19, so to upgrade to that we have to use TLS in our db drivers.

We seem to only use tls for mysql (does this make sense)? Here I'm setting a custom tls config that sets the min version of tls that we support to 1.0 from go1.19's default of 1.2

## Test plan
Have not tested this.


joshma (commented):
I think this looks right, will ask @francescaleung to sign off though - @francescaleung was there a reason why we didn't do this in the first place?

leeweisberger:
> I think this looks right, will ask @francescaleung to sign off though - @francescaleung was there a reason why we didn't do this in the first place?

As an fyi I'm not 100% confident in this so please let me know if we're all not that confident :)

pkg/resources/kinds/mysql.go
```diff
@@ -15,6 +16,12 @@ const SQLDriverMySQL SQLDriver = "mysql"
 
 func init() {
 	resources.RegisterBaseResourceFactory(ResourceKindMySQL, func() resources.Resource { return &MySQLResource{} })
+	if err := mysql.RegisterTLSConfig("skip-verify-tls10", &tls.Config{
+		MinVersion:         tls.VersionTLS10,
+		InsecureSkipVerify: true,
```
francescaleung, line 21: Do we need this? Is it enough to use a min version of TLS 1.0?

francescaleung:
> We seem to only use tls for mysql (does this make sense)?

I think this is true; my recollection is that this is primarily for older versions of MySQL.

pkg/resources/kinds/mysql.go
francescaleung: Does this mean we need a non-`skip-verify` version of this with `MinVersion=tls.VersionTLS10`?

leeweisberger: I can add that. But the option other than skip-verify is false which means no tls and we probably don't need the MinVersion in that case since we aren't even using tls.

francescaleung: oh true

francescaleung:
> I think this looks right, will ask @francescaleung to sign off though - @francescaleung was there a reason why we didn't do this in the first place?

My recollection is that the `tls10default=1` solution was suggested by the Go release notes at the time. @leeweisberger can we remove these lines?

francescaleung (approved):

PR #411, opened 2022-11-15
Send env vars when creating deployment

leeweisberger:
## Description
In bundle discovery, send env vars along with the bundles. Env vars are calculated from `airplane.yaml` and the task definition.

The original plan here was to calculate these in the bundler. However env vars contain secret values that must be resolved. I don't think we should do this in the bundler because we'd have to expose an endpoint to resolve secret values and the bundler doesn't have any sort of advanced authn. Let me know if you have any other thoughts here, or else we'll just send these with the bundle.

## Test plan
I wrote a unit tests


pkg/deploy/discover/path.go
```diff
@@ -64,6 +65,13 @@ func taskPathMetadata(file string, kind build.TaskKind) (TaskPathMetadata, error
...
 	}
+	envVars := make(map[string]build.EnvVarValue)
+	for k, v := range c.EnvVars {
+		envVars[k] = build.EnvVarValue(v)
+	}
+	if len(envVars) == 0 {
+		envVars = nil
+	}
 
 	return TaskPathMetadata{
```
ameyakhare, line 76: I feel like this really shouldnt be the place you do this, this is path-related metadata for the task, not things that are unrelated like env vars

ameyakhare: either change the name, or use a separate function

ameyakhare (changes requested):

ameyakhare (approved):

PR #417, opened 2022-11-16
Fix view bundling

leeweisberger:
## Description
A few fixes discovered while manually testing deploying bundle based views.
- No need to add an extra `/views` directory for the index.html and main.tsx 
- Set base to "" so that assets (built javascript) are hosted relative to the index.html

## Test plan
I manually tested that I can deploy a view with this configuration.


ameyakhare (changes requested):
tests arent passing

ameyakhare (approved):

PR #418, opened 2022-11-16
Login to docker for tests

calufornia:
## Description

We're seeing docker pull request limits — we authenticate with Airplane's docker account to prevent this. Also increases the timeout on Golang tests, as some runs have taken longer than ten minutes.

## Test plan
Ensure golang tests pass in GH workflow


.github/workflows/golang.yml
```diff
@@ -35,6 +35,10 @@ jobs:
     runs-on: buildjet-8vcpu-ubuntu-2004
     steps:
       - uses: actions/checkout@v3
+      - uses: docker/login-action@v2
+        with:
+          username: airplanedevci
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
yolken-airplane, line 41: I think we might also need to set this secret in the project. Let me check...

yolken-airplane: It's not, let me see if I can find it.

yolken-airplane: Ok, should be fixed now!

yolken-airplane (approved):

PR #422, opened 2022-11-17
Don't fill in the default timeout value since it gets set in the API

kellymtrinh:
## Description
We want allow workflows to have a timeout value of 0. Stop setting a default timeout to 3600 on deploy, since timeout validation and defaults are already handled in the API.

This requires a CLI change to ingore diffs where the task yaml doesn't have a timeout defined (but a previous version in the db has it filled out as 3600).

## Test plan
- [x] No timeout specified in task yaml still results in 3600 in the db (it gets filled in from the API)


pkg/deploy/taskdir/definitions/defaults.go
```diff
@@ -73,5 +73,5 @@ func (d DefaultTimeoutDefinition) MarshalJSON() ([]byte, error) {
 }
 
 func (d DefaultTimeoutDefinition) IsZero() bool {
-	return d.Value() == 3600
+	return d.Value() == 0
```
kellymtrinh, line 76: If we don't want to show the diff this would be `	return d.Value() == 0 || d.Value() == 3600` but that feels messy 🥴

pkg/deploy/taskdir/definitions/defaults.go
```diff
@@ -57,7 +57,7 @@ func NewDefaultTimeoutDefinition(value int) DefaultTimeoutDefinition {
 
 func (d DefaultTimeoutDefinition) Value() int {
 	if d.value == 0 {
-		return 3600
+		return 0
 	} else {
 		return d.value
 	}
```
leeweisberger, line 63: This can just be `return d.value`?

pkg/deploy/taskdir/definitions/defaults.go
leeweisberger: Can you be more clear here? Is this going to affect the UI today by showing a diff on every deploy?

kellymtrinh: Yeah it would show a diff for tasks that have no timeout defined at all - not sure that this is what we want to do...
since historically not defining one means it gets filled in as 3600 in the deploy (AND api)

kellymtrinh: I left a screenshot of what this looks like in the PR description

kellymtrinh: Update: we can add a special case to not show the diff if it's not defined but the deployed version has the default here: https://github.com/airplanedev/cli/blob/d8ed8a442c70944f22970fe7236a94a62796dcb2/cmd/airplane/tasks/deploy/deployer.go#L583-L598 in the part that generates the diff?

leeweisberger: Ok you can give that a try. Showing the diff every time a user deploys is something we definitely want to avoid

pkg/deploy/taskdir/definitions/def_0_3_test.go
```diff
@@ -1171,7 +1175,7 @@ func TestDefinitionToUpdateTaskRequest_0_3(t *testing.T) {
 					DisallowSelfApprove: pointers.Bool(false),
 					RequireRequests:     pointers.Bool(false),
 				},
-				Timeout: 3600,
```
kellymtrinh, line 1174: If it's not defined, it won't automatically fill in 3600 when unmarshaling

leeweisberger (approved):

PR #426, opened 2022-11-28
Init inline python tasks without a 0 timeout field

kellymtrinh:
## Description

In https://github.com/airplanedev/lib/pull/422 we stop setting the timeout to 3600 at deploy time if it's not specified (aka 0 value for an int). The API is responsible for setting default timeouts to 3600. 

This fixes python **inline** `tasks init` so it doesn't generate new inline python tasks with a timeout of 0, instead just not having the field by default.

Screen Shot 2022-11-28 at 10 49 49 AM

## Test plan
- [x] Running `airplane tasks init --inline` for python generates a valid task, with no mention of a timeout. Deploying this gets created with the default timeout value of 3600.
- [x] Running `airplane tasks init --from` for an inline task doesn't show the timeout if it's 3600.


pkg/runtime/python/python.go
```diff
@@ -297,9 +297,6 @@ import airplane
     {{- else}}
     allow_self_approvals=False,
     {{- end}}
-    {{- if ne .Timeout 3600}}
```
kellymtrinh, line 300: If you run `airplane tasks init --inline` it currently generates
```

@airplane.task(
    slug="test_python_inline",
    name="test_python_inline",
    require_requests=False,
    allow_self_approvals=True,
)
```

ericliuche (approved):

PR #427, opened 2022-11-28
Use builtins client from runconfig in rest/sql runtime

ameyakhare:
## Description
Use the builtins client passed in using the run config instead of initializing one locally. This way we can use the same builtins client as the dev server in this runtime method.

## Test plan
Builds and executes properly.


pkg/runtime/runtime.go
```diff
@@ -116,6 +117,8 @@ type PrepareRunOptions struct {
 	KindOptions build.KindOptions
 
 	TaskSlug string
+
```
kellymtrinh, line 120: Can you add a comment saying that this is the `Optional builtin client for tasks that need it (such as top level SQL, rest, or builtin) tasks`. 

Also, I just did a search for references for usages of `builtins.NewLocalClient`  and there's one more in `pkg/runtime/builtin/builtin.go`

kellymtrinh (approved):

PR #433, opened 2022-11-30
Support inline tasks in views

leeweisberger:
## Description
Currently you can't define an inline task in a view because executing the task in a node environment fails with browser-specific concepts that are pulled in by the view.

I fixed this for discovery by replacing the document with JSDOM.

We have to do the same thing when builiding locally and remotely.

In order to use esbuild with JSDOM we have to use a simple plugin. In order to use a plugin we have to call esbuild's JS API rather than using the CLI or the go api. I've wanted to switch to this anyway so we can use plugins.

I switched local building to use the new universal shim so that it's more similar to the real prod build

## Test plan
I moved over the whole suite of unit test -> bundle building. I also added tests for the local dev bit.


pkg/build/esbuild.js
```diff
@@ -0,0 +1,41 @@
+const esbuild = require("esbuild");
```
leeweisberger, line 1: This is our new esbuild builder

pkg/build/node.go
```diff
@@ -855,11 +865,6 @@ func nodeBundle(
 		ARG BUILD_NPM_TOKEN
 		RUN [ -z "${BUILD_NPM_RC}" ] || echo "${BUILD_NPM_RC}" > .npmrc
 		RUN [ -z "${BUILD_NPM_TOKEN}" ] || echo "//registry.npmjs.org/:_authToken=${BUILD_NPM_TOKEN}" > .npmrc
-		# qemu (on m1 at least) segfaults while looking up a UID/GID for running
```
leeweisberger, line 858: We're installing this in the shim

pkg/runtime/javascript/javascript.go
```diff
@@ -400,19 +399,7 @@ func (r Runtime) PrepareRun(ctx context.Context, logger logger.Logger, opts runt
...
-	entrypointFunc, _ := opts.KindOptions["entrypointFunc"].(string)
-	shim, err := build.TemplatedNodeShim(build.NodeShimParams{
-		Entrypoint:     filepath.Join("..", entrypoint),
-		EntrypointFunc: entrypointFunc,
-	})
-	if err != nil {
-		return nil, nil, err
-	}
-
+	shim := build.UniversalNodeShim
```
leeweisberger, line 402: We're using the universal shim now. I didn't have to do this, but we might as well use the same shim in local / remote

pkg/runtime/javascript/javascript.go
```diff
@@ -451,41 +438,68 @@ func (r Runtime) PrepareRun(ctx context.Context, logger logger.Logger, opts runt
...
 
-	if len(res.OutputFiles) == 0 {
-		return nil, nil, errors.New("esbuild failed: see logs")
-	}
-
-	return []string{"node", res.OutputFiles[0].Path, string(pv)}, closer, nil
+	entrypointFunc, _ := opts.KindOptions["entrypointFunc"].(string)
+	entrypointExt := filepath.Ext(entrypoint)
+	entrypointJS := strings.TrimSuffix(entrypoint, entrypointExt) + ".js"
+	return []string{"node", filepath.Join(taskDir, "dist", "shim.js"), filepath.Join(taskDir, "dist", entrypointJS), entrypointFunc, string(pv)}, closer, nil
```
leeweisberger, line 502: Now this is more consistent with how tasks are executed in the bundle build world

pkg/build/node_test.go
```diff
@@ -251,6 +251,570 @@ func TestInlineConfiguredTasks(t *testing.T) {
 	ctx := context.Background()
 
 	tests := []Test{
+		{
```
leeweisberger, line 254: I ported all hte tests over from the non-bundle-build test suite

pkg/build/node.go
```diff
@@ -765,18 +774,19 @@ func nodeBundle(
...
 		for _, dep := range deps {
-			flags = append(flags, fmt.Sprintf("--external:%s", dep))
+			flags = append(flags, fmt.Sprintf(`"%s"`, dep))
 		}
 		if isWorkflow {
 			// Even if these are imported, we need to mark the root packages
 			// as external for esbuild to work properly. Esbuild doesn't
 			// care about repeats, so no need to dedupe.
-			flags = append(flags, "--external:@temporalio", "--external:@swc")
+			flags = append(flags, `"@temporalio"`, `"--external:@swc"`)
```
calufornia, line 786: How come we still mark `swc` as external?

pkg/runtime/javascript/javascript.go
```diff
@@ -451,41 +438,68 @@ func (r Runtime) PrepareRun(ctx context.Context, logger logger.Logger, opts runt
...
+	if err := os.WriteFile(filepath.Join(taskDir, "esbuild.js"), []byte(build.Esbuild), 0644); err != nil {
+		return nil, nil, errors.Wrap(err, "writing esbuild file")
+	}
+	// First build the shim.
+	cmd = exec.CommandContext(ctx,
+		"node",
+		filepath.Join(taskDir, "esbuild.js"),
+		fmt.Sprintf(`["%s"]`, filepath.Join(taskDir, "shim.js")),
+		"node"+build.GetNodeVersion(opts.KindOptions),
+		fmt.Sprintf(`["%s"]`, strings.Join(external, ", ")),
```
calufornia, line 458: This shouldn't be quoted, right? Since the external deps are already quoted so it would end up being `[""a", "b""]`

pkg/runtime/javascript/javascript.go
```diff
@@ -451,41 +438,68 @@ func (r Runtime) PrepareRun(ctx context.Context, logger logger.Logger, opts runt
...
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "entrypoint is not within the task root")
 	}
-	logger.Debug("Compiled JS in %s", time.Since(start).String())
+	cmd = exec.CommandContext(ctx,
+		"node",
+		filepath.Join(taskDir, "esbuild.js"),
+		fmt.Sprintf(`["%s"]`, filepath.Join(root, entrypoint)),
+		"node"+build.GetNodeVersion(opts.KindOptions),
+		fmt.Sprintf(`["%s"]`, strings.Join(external, ", ")),
```
calufornia, line 479: Same here

pkg/runtime/javascript/javascript.go
```diff
@@ -451,41 +438,68 @@ func (r Runtime) PrepareRun(ctx context.Context, logger logger.Logger, opts runt
...
+	)
+	cmd.Dir = filepath.Join(root, ".airplane")
+	logger.Debug("Running %s (in %s)", strings.Join(cmd.Args, " "), root)
+	out, err = cmd.CombinedOutput()
+	if err != nil {
+		logger.Log(strings.TrimSpace(string(out)))
+		return nil, nil, errors.New("failed to build task")
+	}
+
+	logger.Debug("Built JS in %s", time.Since(start).String())
```
calufornia, line 492: Do you know how much longer/shorter it takes to prepare the run using the universal node shim?

calufornia (commented):
Runtime changes mostly look good to me, a couple comments

pkg/runtime/javascript/javascript.go
leeweisberger: I didn't measure. But the universal shim doesn't add anything that would add latency except that the shim has to be bundled which should be very quick

calufornia (approved):

ameyakhare (changes requested):
leave this one out for tomorrow

leeweisberger:
Closing in favor of smaller prs

PR #435, opened 2022-11-30
Add ScrubCalculatedFields to Resource interface

calufornia:
## Description

Adds the ability to scrub calculated fields from resources, the inverse operation to `Calculate()`. We'll call this method before saving resources to the dev config file.

## Test plan
Test with CLI changes that call `ScrubCalculatedFields()`


calufornia:
Side note: I don't really like the empty `ScrubCalculatedFields` methods, and we should probably just have a default implementation on the `BaseResource` (same with `Calculate`).

pkg/resources/resource.go
```diff
@@ -99,16 +99,15 @@ type Resource interface {
...
 	// Validate returns an error if the resource is invalid.
 	Validate() error
 	// Kind returns the ResourceKind associated with this resource.
 	// Deprecated: Use GetKind() instead
 	Kind() ResourceKind // TODO: Remove
 	// String returns a string representation of this resource.
 	String() string
-	// ID returns the resource's ID.
-	// Deprecated: Use GetID() instead
-	ID() string // TODO: Remove
```
calufornia, line 111: This was unused

pkg/resources/kinds/rest.go
```diff
@@ -220,3 +227,6 @@ func (a *RESTAuthBasic) calculate() error {
 	}
 	return nil
 }
+
+// Do nothing — we can't determine if the Authorization header is calculated or defined by the user.
+func (a *RESTAuthBasic) scrubCalculatedFields() {}
```
francescaleung, line 232: headers in auth are always calculated--this should scrub the auth headers.

If folks want custom Authorization headers they shouldn't be using AuthBasic, just throw it in the regular REST headers.

pkg/resources/resource.go
francescaleung: :tada:

francescaleung (approved):

PR #436, opened 2022-11-30
Omit empty optional and calculated resource fields

calufornia:
## Description

Second part in hiding calculated fields from the dev yaml file — omit those fields if they are empty. Primarily applies to calculated fields, but we also omit empty, optional fields.

## Test plan
Test in conjunction with CLI changes


pkg/resources/kinds/bigquery.go
```diff
@@ -22,11 +22,11 @@ type BigQueryResource struct {
 	resources.BaseResource `mapstructure:",squash" yaml:",inline"`
 
 	Credentials    string `json:"credentials" mapstructure:"credentials"`
-	RawCredentials string `json:"rawCredentials" mapstructure:"rawCredentials"`
+	RawCredentials string `json:"rawCredentials,omitempty" mapstructure:"rawCredentials"`
```
francescaleung, line 25: I have a PR I really need to merge in that changes web to send in RawCredentials & not Credentials. My gut feel here is that raw credentials shouldn't need this tag but the limbo is weird so I'll defer to you.

francescaleung (approved):

pkg/resources/kinds/bigquery.go
calufornia: SGTM - I won't scrub raw credentials either, since it's more human-readable (and editable) than credentials

PR #437, opened 2022-12-01
Fix node entrypoint for bundle builds

leeweisberger:
## Description
The entrypoint should be the build version of the entrypoint which is always js. For non-inline tasks, since the definition is created by the user and could point to a ts file, we need to update the entrypoint to use the built js file.

Since inline task defs are parsed from the built js file, they dont have this problem.

## Test plan
Tested manually and ensured we can execute the task!


pkg/build/python_test.go
```diff
@@ -189,22 +189,22 @@ func TestPythonBundleBuilder(t *testing.T) {
 				Version: BuildTypeVersionPython37,
 			},
 			FilesToDiscover: []string{
-				"rootInlineTask.airplane.py",
-				"subfolder/subfolderInlineTask.airplane.py",
+				"rootInlineTask_airplane.py",
```
leeweisberger, line 192: This is unrelated

pkg/build/builder_test.go
```diff
@@ -134,7 +134,7 @@ func RunTests(tt *testing.T, ctx context.Context, tests []Test) {
 							out := runTask(t, ctx, client, runTaskConfig{
 								Image:       resp.ImageURL,
 								ParamValues: test.ParamValues,
-								Entrypoint:  []string{"python", "/airplane/.airplane/shim.py", path.Join("/airplane/", testRun.RelEntrypoint), testRun.ExportName},
+								Entrypoint:  []string{"python", "/airplane/.airplane/dist/shim.py", path.Join("/airplane/.airplane/", testRun.RelEntrypoint), testRun.ExportName},
```
leeweisberger, line 137: We're storing the shim under `dist` and the final files under `/airplane/.airplane`. Same thing as node

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -158,10 +159,14 @@ func (d *NodeDefinition_0_3) fillInUpdateTaskRequest(ctx context.Context, client
 	req.Env = d.EnvVars
 	if forBundle {
 		entrypointFunc, _ := bc["entrypointFunc"].(string)
+		entrypoint := path.Join("/airplane/.airplane/", bc["entrypoint"].(string))
+		// Ensure that the entyrpoint is a .js file.
+		entrypointExt := filepath.Ext(entrypoint)
+		entrypoint = strings.TrimSuffix(entrypoint, entrypointExt) + ".js"
```
leeweisberger, line 165: Here's the fix for running bundled node tasks

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -272,8 +277,8 @@ func (d *PythonDefinition_0_3) fillInUpdateTaskRequest(ctx context.Context, clie
 		entrypointFunc, _ := bc["entrypointFunc"].(string)
 		req.Command = []string{"python"}
 		req.Arguments = []string{
-			"/airplane/.airplane/shim.py",
-			path.Join("/airplane/", bc["entrypoint"].(string)),
+			"/airplane/.airplane/dist/shim.py",
+			path.Join("/airplane/.airplane", bc["entrypoint"].(string)),
```
leeweisberger, line 281: And updating python to match the build change

ameyakhare (approved):

PR #441, opened 2022-12-02
Update node parser to support convertToDef

ericliuche:
## Description
In https://github.com/airplanedev/js-sdk/pull/214, we added a `convertToDef` helper function to extract out the config parsing logic. This allows us to more easily make changes to the config -> def conversion without having to go through lib

## Test plan
Ran locally


colinking (changes requested):
(see comments on js-sdk)

ameyakhare (changes requested):
requesting changes to clear from queue

colinking:
@ericliuche you will need to move this PR to `airplanedev/cli`.

PR #444, opened 2022-12-02
Reconfigure how airplane.yaml is organized. Add install options

leeweisberger:
## Description
This PR contains two changes:

1. Reconfigure airplane.yaml file to nest options under `javascript` and `python`.

Before, all fields in the `airplane.yaml` were top level. This meant that an Airplane project that contained multiple languages (python and node) had to share all configuration. This doesn't make sense for things like install commands and might differ for other options.

Moreover, some of these options don't make sense for certain tasks. e.g. sql tasks don't abide by base or envvars.

Although this makes the config file slightly more verbose, I think it's more correct and will avoid issues in the future.

Before
```yaml
nodeVersion: "18"
base: "slim"
```
After
```yaml
javascript:
  nodeVersion: "18"
  base: "slim"
  preinstall: echo "preinstall! javascript"
python:
  base: "full"
  preinstall: echo "preinstall python"
```

2. Add preinstall/postinstall command options to the airplane.yaml. This is an additional way to configure these. Add install to javascript (we don't currently support this for python, but we could add).

## Test plan
Unit tests in the builder ensure that we can successfully parse the airplane config and read in options.


pkg/deploy/config/taskenv.go
```diff
@@ -0,0 +1,77 @@
+package config
+
+// This file is copied from pkg/api/client.go to avoid circular dependency issues.
```
ericliuche, line 3: sad

ericliuche (approved):

pkg/deploy/config/taskenv.go
leeweisberger: I tried a few ways to do this but all of them required huge changes

PR #445, opened 2022-12-02
Set env vars on task defn during discovery

leeweisberger:
## Description
Ensure that the task defn is the source of truth for env vars. Add in env vars from the airplane.yaml config file.

Without this change, we have to keep looking for and reading in the config file i.e. during local dev. With this change downstream code remains unchanged

## Test plan
Unit tested


pkg/deploy/discover/discover_test.go
```diff
@@ -715,6 +716,80 @@ func TestDiscover(t *testing.T) {
 			},
 			defnFilePaths: []string{fixturesPath + "/taskInlineEsm/codeOnlyTask.airplane.ts"},
 		},
+		{
```
calufornia, line 719: Can you add a test for the override behavior? i.e. preference of definition env vars > airplane.yaml env vars

calufornia (approved):

PR #446, opened 2022-12-02
Implement bundle builds for workflows

ameyakhare:
Resolves TSK-999

## Description
Implements bundle builds for workflows.

## Test plan
Deployed and ran a locally built workflow successfully.


pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -158,12 +158,20 @@ func (d *NodeDefinition_0_3) fillInUpdateTaskRequest(ctx context.Context, client
...
 		entrypointFunc, _ := bc["entrypointFunc"].(string)
-		req.Command = []string{"node"}
-		req.Arguments = []string{
-			"/airplane/.airplane/dist/universal-shim.js",
-			path.Join("/airplane/.airplane/", bc["entrypoint"].(string)),
-			entrypointFunc,
-			"{{JSON.stringify(params)}}",
+		if req.Runtime == build.TaskRuntimeWorkflow {
+			req.Arguments = []string{
+				"{{JSON.stringify(params)}}",
```
ameyakhare, line 163: this needed to change and is an implementation detail that we should clean up.

workflow args are passed from args[1] and args[2], but params must be in args[0]

pkg/build/node.go
```diff
@@ -839,7 +869,9 @@ func nodeBundle(
 	// Down the road, we may want to give customers more control over this build process
 	// in which case we could introduce an extra step for performing build commands.
 	return applyTemplate(heredoc.Doc(`
-		FROM {{.Base}}
+		FROM {{.Base}} as base
+		ENV NODE_ENV=production
+		WORKDIR /airplane{{.Workdir}}
```
leeweisberger, line 874: We can actually remove the 
```
RUN apt-get update && export...
```
command. This is only needed at runtime, not install time

pkg/build/workflow/universal-workflow-shim.js
```diff
@@ -0,0 +1,66 @@
...
+import {
+  proxySinks,
+  isCancellation,
+  CancellationScope,
+} from "@temporalio/workflow";
+import { registerWorkflowRuntime } from "@airplane/workflow-runtime/internal";
+
+const taskImports = {
+  {{- range $taskImport := .TaskImports}}
+  "{{$taskImport.CompiledFile}}": require("../{{$taskImport.UserFile}}"),
```
leeweisberger, line 11: So a relative require works here... You're sure that the same relative require doesn't work if we just require it dynamically? i.e. can we do this without the lookup table? 

So the same exact code I handed over to you where the entrypoint is preceded by a `../`?

pkg/deploy/taskdir/definitions/def_0_3.go
leeweisberger: We need params here? Is this being used? Can we just delete the first element of the array and change the aiport code to reading the entrypoint and entrypointfunc from the first and second element instead of the second and third?

leeweisberger (commented):
I'm fine with this, but did you test this without the lookup table but with the entrypoint?

pkg/build/node.go
leeweisberger: Put this in the base image and remove from everywhere else

pkg/build/workflow/universal-workflow-shim.js
leeweisberger: We tried doing this and got errors

pkg/deploy/taskdir/definitions/def_0_3.go
leeweisberger: We are literally passing these args to the workflow

leeweisberger (approved):

PR #461, opened 2022-12-06
Remove min version wf warning

leeweisberger:
## Description
Now that we are building for tasks and wfs using the same dockerfile, we don't know if a workflow should be built at this point. Remove this error so it doesn't pop up when users are just building tasks.


leeweisberger:
The wfs team said this was ok in slack!

ameyakhare (approved):

PR #465, opened 2022-12-08
Support installing view dependencies from a private repo

leeweisberger:
## Description
Support setting the .npmrc file the same way we do for tasks. We pull build-time env vars from the new airplane.yaml configuration file.

This will only work once bundle based deployments ship which should be in the next day or two.

## Test plan
I didn't test this e2e, but I tested that we are setting the args and passing them into the build, so no reason this shouldn't work.
image


calufornia (approved):
Should we also include this in the [non-bundle build template](https://github.com/airplanedev/lib/blob/bba2d0db5d9d0322dc4b25e49d8f009c043138c0/pkg/build/view.go#L133) until we've turned on bundle builds for all teams?

leeweisberger:
> Should we also include this in the [non-bundle build template](https://github.com/airplanedev/lib/blob/bba2d0db5d9d0322dc4b25e49d8f009c043138c0/pkg/build/view.go#L133) until we've turned on bundle builds for all teams?

It's not going to work there because there is no way to set env vars in views currently!

PR #467, opened 2022-12-08
Don't parse node version or base from inline config tasks

leeweisberger:
## Description
Stop parsing node version and base from inline config tasks.
1. These are being removed from the sdk
2. Parsing these but ignoring them when creating bundles can cause tasks to not be discovered. The bundle is created with an empty node version but on parsing we try to compare that to the node version parsed out here.


pkg/deploy/discover/parser/node/parser.ts
```diff
@@ -110,10 +110,8 @@ const extractTaskConfigs = (files: string[]): AirplaneConfigs => {
             parameters: params,
             entrypointFunc: exportName,
             node: {
```
ameyakhare, line 112: please remove from NodeDef above as well

ameyakhare (changes requested):

ameyakhare (approved):

PR #469, opened 2022-12-08
Re-enable renovate for fixtures

leeweisberger:
Re-enable renovate on fixtures packages. Even though it's annoying, it's best to keep these up to date.


colinking (approved):
The fixtures that were causing issues weren't under a `fixtures` directory: https://github.com/airplanedev/lib/commit/0b4a2f0557a3143a1c1055ad1a780f8e7be5c153

Is there something else preventing Renovate from updating those fixtures?

Also there are a few fixtures that should not be updated, e.g. `javascript/workflowold`. Can you add ignore paths for those?

leeweisberger:
> The fixtures that were causing issues weren't under a `fixtures` directory: [0b4a2f0](https://github.com/airplanedev/lib/commit/0b4a2f0557a3143a1c1055ad1a780f8e7be5c153)
> 
> Is there something else preventing Renovate from updating those fixtures?
> 
> Also there are a few fixtures that should not be updated, e.g. `javascript/workflowold`. Can you add ignore paths for those?

Yeah it looks like we had a renovate pr that was just around for a long time and didn't get merged. Going to merge this anyways. We also deleted the workflowold so no need to ignore, but if we add something like this in the future we can ignore

PR #472, opened 2022-12-09
Add default appURL in explain error messages

kellymtrinh:
## Description
Fixes this error where the CLI would print out
```
Error: Resource with slug "launchdarkly" does not exist

Follow the URL below to create the resource:
/settings/resources/new
```
With a bad URL.

I'm adding a default here because the error shouldn't assume that all clients always remember to pass in a host, or have access to a host. The discoverer, which was what was setting the empty url in the error message, doesn't have access to the host and is a few layers deep in airport and CLI 😬 . The other usages of `ResourceMissingError` and TaskMissingError in CLI and airport pass in an appURL.

Also updated a few instances where we return a different error message instead of the resource missing error. 

## Test plan
- [x] Now the logic that validates resources happens in the builder and not in the CLI, which doesn't print the ExplainError(), so this is kind of a non issue now :/ : 
This is what the error message looks like now:
Screenshot 2022-12-08 at 4 36 16 PM

Before this, the error from the builder was `unknown resource: "demo123"`

CLI printed it here previously https://github.com/airplanedev/cli/blob/82796c6e648f6537b395db0bbe1915c85b644575/cmd/airplane/main.go#L38


pkg/api/errors.go
```diff
@@ -2,6 +2,21 @@ package api
 
 import "fmt"
 
+const (
+	DefaultAppURL = "https://app.airplane.dev"
+)
+
+func getAppURL(appURL string) string {
+	if len(appURL) > 0 {
```
leeweisberger, line 10: if appURL != "" ?

leeweisberger (approved):

PR #473, opened 2022-12-09
Always defualt views to the slim base

leeweisberger:
## Description
All views, inline or not, default to the slim base. Imoved this into a helper function rather than directly in the discoverer to airport can use the same helper function.

As written, the code ties the base of the view to the base of other JS tasks in the project as specified in the `airplane.yaml`. I **think** I like this as it gives users an out (they can use the full base for views) and if we do decide to build tasks and views together, they'll be of the same base.

In the future we could make a `views` section of airplane.yaml where you can specify the views base separately from your javascript tasks. I don't really think this is going to be a big deal though, and we could always do this in the future.

## Test plan
Tested manually e2e


pkg/deploy/discover/build_context.go
```diff
@@ -6,11 +6,12 @@ import (
...
 	"github.com/airplanedev/lib/pkg/deploy/config"
 	"github.com/airplanedev/lib/pkg/runtime"
+	"github.com/airplanedev/lib/pkg/runtime/javascript"
 	"github.com/airplanedev/lib/pkg/utils/fsx"
 )
 
-// taskBuildContext gets the build context for a task.
-func taskBuildContext(taskroot string, taskRuntime runtime.Interface) (build.BuildContext, error) {
+// TaskBuildContext gets the build context for a task.
+func TaskBuildContext(taskroot string, taskRuntime runtime.Interface) (build.BuildContext, error) {
```
leeweisberger, line 14: Exporting this so airport can use it

pkg/deploy/taskdir/definitions/def_0_3.go
```diff
@@ -378,6 +378,9 @@ func (d *PythonDefinition_0_3) getBuildType() (build.BuildType, build.BuildTypeV
 }
 
 func (d *PythonDefinition_0_3) SetBuildVersionBase(v build.BuildTypeVersion, b build.BuildBase) {
+	if d.Base == "" {
+		d.Base = b
+	}
```
leeweisberger, line 383: This was a bug - we weren't setting the base for python so we were never builiding slim!

pkg/build/view.go
```diff
@@ -166,8 +166,9 @@ func viewBundle(root string, buildContext BuildContext, options KindOptions, fil
 		apiHost = "https://" + apiHost
 	}
 
+	useSlimImage := buildContext.Base == BuildBaseSlim
 	nodeVersion := GetNodeVersion(options)
-	base, err := getBaseNodeImage(nodeVersion, false)
+	base, err := getBaseNodeImage(nodeVersion, useSlimImage)
```
leeweisberger, line 171: We were never using a slim image for views!

pkg/deploy/taskdir/definitions/view_def.go
```diff
@@ -7,18 +7,20 @@ import (
...
-	Entrypoint   string      `json:"entrypoint"`
-	EnvVars      api.EnvVars `json:"envVars,omitempty"`
-	DefnFilePath string      `json:"defnFilePath,omitempty"`
+	Name         string          `json:"name"`
+	Slug         string          `json:"slug"`
+	Description  string          `json:"description,omitempty"`
+	Entrypoint   string          `json:"entrypoint"`
+	EnvVars      api.EnvVars     `json:"envVars,omitempty"`
+	DefnFilePath string          `json:"defnFilePath,omitempty"`
+	Base         build.BuildBase `json:"base,omitempty"`
```
ameyakhare, line 23: why is this part of the definition?

ameyakhare (approved):

pkg/deploy/taskdir/definitions/view_def.go
leeweisberger: It's basically so we don't have to keep re-figuring this out. You discover a view. Part of that discovery involves calculating the base. Now you don't have to calculate the base on your own. We do the same thing for tasks fwiw

PR #475, opened 2022-12-09
Improve error message when discoverig inline tasks and views

leeweisberger:
## Description
When we encounter an error while discovering inline tasks/views, we used to just spit out an unformated esbuild string.

Now we pull out the actual message, remove references to esbuild, and hint at the user that they may have forgotten to install dependency(s)

Before
image

After
image

## Test plan
Manually tested. See screenshots


pkg/deploy/discover/build.go
```diff
@@ -68,13 +69,21 @@ func esbuildUserFiles(rootDir string) error {
...
+			msg = fmt.Sprintf("%s. Did you forget to install a dependency?", msg)
+		}
+		errMsgs = append(errMsgs, msg)
 	}
 
 	if len(res.OutputFiles) == 0 {
-		return errors.New("esbuild failed to produce output files")
+		msg := "failed to build code"
+		if len(errMsgs) > 0 {
+			msg = strings.Join(errMsgs, ", ")
```
calufornia, line 84: Can we do a new line here? I think that makes it more clear they're separate errors

calufornia: Should we also append the `"failed to build code"` message here?

calufornia (approved):

pkg/deploy/discover/build.go
leeweisberger: That's just a backup message. They still get `Unable to build task` prepended, so I don't think they need both

PR #477, opened 2022-12-10
Target files is now the files that should be built

leeweisberger:
## Description
Change the definition of target files a bit.

Previously, we sent a bundle up without specifying all of the files that should be included in the bundle. We expected the server to run the same bundle discovery logic and decide which files should be included in the bundle.

However, this requires the server's logic to stay in sync with the client's logic which will be a nightmare since we can't ensure all users are using the latest cli.

We either need to (a) Ensure that the logic is only on the server or (b) Ensure that the logic is only on the client and the client is sending additional information. We're not counting out (a), but it has some technical and perf limitations (we need to download user code on the server, run bundle discovery, and then kick off builds). This is an implementation of (b) where the additional information is all of the files that the client expected to be in the bundle. The server will listen to this information rather than trying to calculate on its own.

## Test plan
We have a good unit test suite. The only thing I had to update was to change targetFiles to full paths to the files. Nothing else had to change.

I also ran this e2e (with server changes) against my suite of manual tests and it worked as expected.


pkg/deploy/bundlediscover/discover.go
```diff
@@ -81,103 +81,100 @@ func (d *Discoverer) discoverHelper(ctx context.Context, paths ...string) ([]Bun
...
-				}
-				if bundlePath == "" {
-					// This file is not an Airplane task.
-					continue
-				}
-
-				b := Bundle{
-					RootPath:     bundlePath,
-					BuildContext: buildContext,
+			err := filepath.WalkDir(p, func(path string, entry fs.DirEntry, err error) error {
```
leeweisberger, line 96: Using filePath.Walkdir instead of our own recursion. This makes things a lot easier!

pkg/deploy/bundlediscover/discover.go
```diff
@@ -81,103 +81,100 @@ func (d *Discoverer) discoverHelper(ctx context.Context, paths ...string) ([]Bun
...
+			if err != nil {
+				return nil, err
 			}
 		}
 	}
 
 	return bundles, nil
 }
 
+func (d *Discoverer) getBundlesForFile(ctx context.Context, path string) ([]Bundle, error) {
```
leeweisberger, line 122: I refactored this function out with no changes

ameyakhare (approved):

PR #482, opened 2022-12-15
Fix shell builds where base image has entrypoint

leeweisberger:
## Description
If the user supplied Dockerfile has an entrypoint, this entrypoint will be executed instead of our custom CMD/ARGS. We "unset" the entrypoint so it doesn't interfere.

I updated our testing infra a bit to reflect reality - we are using CMD not entrypoint. Without this I couldn't test.

## Test plan
I wrote a unit test that exactly replicated a user's issue. I then simplified it a bit.


pkg/build/shell.go
```diff
@@ -59,6 +59,8 @@ func shellBundle(root string) (string, error) {
 		RUN mkdir -p .airplane && {{.InlineShim}} > .airplane/shim.sh
 
 		COPY --chmod=700 . .
+		# Set an empty entrypoint to override any entrypoints that may be set in the base image.
+		ENTRYPOINT []
```
colinking, line 63: And `CMD` is _always_ set by us, so that's always overwritten, right?

pkg/build/bundle_builder.go
```diff
@@ -188,6 +188,9 @@ func (b *BundleBuilder) Build(ctx context.Context, bundleBuildID, version string
 
 		var resp controlapi.StatusResponse
 
+		if msg.ErrorMessage != "" {
+			return nil, errors.New(msg.ErrorMessage)
```
colinking, line 192: maybe wrap this error message? I'm assuming this'll be a random buildkit error msg?

colinking (approved):

pkg/build/shell.go
leeweisberger: CMD is always set by us. That's done here: https://github.com/airplanedev/lib/blob/main/pkg/deploy/taskdir/definitions/def_0_3.go#L398

PR #484, opened 2022-12-20
Support python 3.11

leeweisberger:
## Description
Support python versions. Started by adding 3.11 and we can expand by supporting additional python versions once this infra is in.

## Test plan
Manually tested that I can init a python task and deploy at different versions


pkg/build/versions.json
```diff
@@ -49,13 +49,33 @@
   "python": {
     "3": {
       "image": "registry.hub.docker.com/library/python",
-      "tag": "3.10.8-bullseye",
-      "digest": "sha256:eef39ed128b235c95c723eabe2de05670ba87f3273cc784effe4c3d9d0847c09"
+      "tag": "3.10.9-bullseye",
```
leeweisberger, line 52: I also updated python 3.10 to latest

pkg/deploy/config/config.go
```diff
@@ -31,6 +31,7 @@ type JavaScriptConfig struct {
 
 type PythonConfig struct {
 	Base        string  `yaml:"base,omitempty" json:"base,omitempty"`
+	Version     string  `yaml:"version,omitempty" json:"version,omitempty"`
```
leeweisberger, line 34: The airplane.yaml file works with python version

pkg/deploy/config/schema.json
```diff
@@ -40,6 +40,10 @@
           "enum": ["", "full", "slim"],
           "default": ""
         },
+        "version": {
+          "description": "The version of Python to use.",
+          "enum": ["3.10", "3.11"]
+        },
```
leeweisberger, line 46: The defn file can specify a python version

pkg/deploy/taskdir/definitions/fixtures/python.task.yaml
```diff
@@ -43,6 +43,9 @@ python:
   # absolute or relative to the location of the definition file.
   entrypoint: my_task.py
 
+  # The version of Python to use. Valid values: "3.10", "3.11". Defaults to "3.10".
+  # version: "3.10"
```
leeweisberger, line 47: By default, don't set the version in the task defn. We're trying to push people to set it in their airiplane.yaml which will get auto-created when you init a task and have the default python version set by default. This is different from node. Thoughts?

pkg/build/versions.json
```diff
@@ -49,13 +49,33 @@
...
+      "digest": "sha256:6e65ad59b36c353d9456e5ca5fd319c672775a786208dce565e25adc1087db86"
```
ericliuche, line 53: according to python's versioning scheme: https://devguide.python.org/developer-workflow/development-cycle/

```
Also, a general rule for maintenance branches is that compatibility must not be broken at any point between sibling micro releases (3.5.1, 3.5.2, etc.). For both rules, only rare exceptions are accepted and must be discussed first.
```

it technically is possible that micro version updates break compatibility. lambda (https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html) doesn't seem to allow you to specify the micro version, but i'm not sure how we should handle bumping it... do you have any thoughts here? how do we handle this for javascript

pkg/build/versions.json
```diff
@@ -49,13 +49,33 @@
   "python": {
     "3": {
       "image": "registry.hub.docker.com/library/python",
-      "tag": "3.10.8-bullseye",
```
ericliuche, line 52: should we might as well add 3.7, 3.8, 3.9? our sdk technically supports 3.7+ (https://github.com/airplanedev/python-sdk/blob/main/pyproject.toml#L14). also lambda supports 3.7+

pkg/build/versions.json
ericliuche, line 52: also a bit unrelated, but we should probably document somewhere exactly which images we're using

pkg/deploy/taskdir/definitions/fixtures/python.task.yaml
ericliuche: should we just not add it at all in this file? i thought setting the version via task defn wasn't supported anymore. my understanding of the behavior was if you don't have an airplane.yaml or don't set the python version, we'd default everything to 3.10 and if you set it, then we respect the version you chose

leeweisberger: You can set it in the defn in JavaScript. You can't set it using an inline task. This would just be keeping parity with JS... I don't think people should really be using this.... I'm not against removing it... I'll remove it

pkg/build/versions.json
leeweisberger: Added

pkg/deploy/config/schema.json
ericliuche, line 46: add 3.7-3.9

ericliuche (approved):

PR #493, opened 2023-01-03
Use esbuild via JS sdk

leeweisberger:
## Description

Rather than using esbuild via the CLI, use the JS sdk. This is pre-work to support esbuild plugins which we need to solve a few issues.

## Test plan
I updated all of the unit tests to run in bundle mode. They all continue to pass with this change


pkg/build/node_test.go
```diff
@@ -260,6 +260,592 @@ func TestNodeBundleBuilder(t *testing.T) {
 	ctx := context.Background()
 
 	tests := []Test{
+		{
```
leeweisberger, line 263: I moved all of the normal tests into the bundle test suite. No additional changes were made.

ameyakhare (approved):
nice

PR #494, opened 2023-01-03
Support task in view file

leeweisberger:
## Description
Currently you can't define an inline task in a view because executing the task in a node environment fails with browser-specific concepts that are pulled in by the view.

I fixed this for discovery by replacing the document with JSDOM.

We have to do the same thing when builiding locally and remotely.

## Test plan
I added a unit test for a task defined in a view. It failed before and succeeds after this change.


pkg/deploy/discover/parser/node/build.ts
```diff
@@ -4,24 +4,18 @@ import esbuild from "esbuild";
...
-          'const syncWorkerFile = require.resolve ? require.resolve("./xhr-sync-worker.js") : null;',
-          `const syncWorkerFile = "${require.resolve(
-            "jsdom/lib/jsdom/living/xhr/xhr-sync-worker.js"
-          )}";`
-        );
-
-        return { contents, loader: "js" };
-      }
-    );
+    build.onLoad({ filter: /XMLHttpRequest-impl\.js$/ }, async (args) => {
```
leeweisberger, line 7: This is just lint

ameyakhare (approved):

PR #496, opened 2023-01-03
Move to latest version of gosnowflake

leeweisberger:
## Description
We pinned gosnowflake due to an issue with a dependency they were using breaking our releases.

They pinned this dependency themselves in the latest release: https://github.com/snowflakedb/gosnowflake/pull/689

We should be able to return to the latest release of gosnowflake.


kellymtrinh (approved):
If pinning this version in CLI passes the release build step LGTM

PR #500, opened 2023-01-06
Enable windows support for builtins

calufornia:
## Description

Add `"windows"` to the list of supported operating systems. The binary that is produced for windows is slightly different (ends in `".exe"`), so we also add logic to handle that.

## Test plan
Build CLI and test in windows


pkg/builtins/local_client.go
```diff
@@ -18,7 +18,7 @@ import (
 	"google.golang.org/api/option"
 )
 
-var supportedOS = map[string]bool{"darwin": true, "linux": true}
+var supportedOS = map[string]bool{"darwin": true, "linux": true, "windows": true}
```
kellymtrinh, line 21: Are both windows amd64 and arm64 supported?

kellymtrinh (approved):

pkg/builtins/local_client.go
calufornia: Hmm good point—only amd64 for now (I don't expect to see much of the latter). I'll change how we check supported OS/arch pairs

PR #501, opened 2023-01-06
Add core API client

colinking:
## Description

This PR adds an `api/http` package which includes a "core" API client implementation that can be reused by our two API client implementations (`cli` and `airport`). This core logic performs exponential retries with idempotency keys.

For usage, see:
- CLI: https://github.com/airplanedev/cli/pull/892
- Airport: https://github.com/airplanedev/airport/pull/3877

Resolves TSK-936

## Test plan

See unit tests.


colinking:
* main
  * **PR #501** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/501" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/501?utm_source=stack-comment).

pkg/api/http/client.go
```diff
@@ -0,0 +1,325 @@
...
+	Timeout time.Duration
+
+	// retryWaitMin overrides the minimum wait time between retries. Used for testing purposes only.
+	retryWaitMin time.Duration
+	// retryWaitMax overrides the maximum wait time between retries. Used for testing purposes only.
+	retryWaitMax time.Duration
+}
+
+// NewClient creates a new Airplane HTTP client that can be used to issue HTTP requests.
+func NewClient(opts ClientOpts) Client {
```
colinking, line 52: This client does not set `User-Agent` like we do in the JS SDK. I'd like to add that, but as-is, the Client does not have any information on teamID/runID/version/etc. It seems more clean for airport/CLI/etc. to generate it instead.

yolken-airplane: Should we make `UserAgent` a required header to ensure clients always set it?

colinking: I like it. Will do.

pkg/api/http/client.go
```diff
@@ -0,0 +1,359 @@
...
+		httpreq.Header.Set(k, v)
+	}
+	if opts.UserAgent != "" {
+		httpreq.Header.Set("User-Agent", opts.UserAgent)
+	}
+
+	// Set an Idempotency-Key header. This will ensure we can safely retry all HTTP requests.
+	switch method {
+	// Idempotency keys are not necessary for idempotent-by-definition methods, such as GET/DELETE.
+	case http.MethodPost, http.MethodPatch:
```
yolken-airplane, line 291: Should we have an option to skip the idempotency key even if the method is POST or PATCH?

I noticed in reviewing the API changes that there are a few POST calls we're making that really could be GETs, but I think we're using the former so we can put the query params in the body (e.g., `/v0/agents/getDesiredRunnerState`). Also, there are some like the heartbeats where we really don't care about idempotency (they should probably be PUTs instead of POSTs). Given the high frequency of these calls and the extra overhead of having the idempotency keys, it might be easier and faster to just skip them here.

yolken-airplane: The `getDesiredRunnerState` responses can also be super large.

colinking: I'd prefer we not make Idempotency-Keys optional, as that complicates the behavior for clients. What if we configure that server-side? We can track a list of API endpoints that are side-effect free (in which case, we ignore Idempotency-Keys). For the heartbeat endpoints, I can migrate those to `PUT` before merging this PR.

yolken-airplane: Sure, as long as the responses of `getDesiredRunnerState` aren't saved in the DB, that's fine by me.

colinking: I'll extract both into separate PRs. Here's the first: https://github.com/airplanedev/airport/pull/3891

colinking: Just updated this PR to include client methods for `PUT` requests.

yolken-airplane (approved):

PR #503, opened 2023-01-09
Add Get{Bundle,}BuildInstructions for Python

francescaleung:
## Description

Pulls out build instructions for Python into a more structured format that can either be applied directly or via Dockerfile.

Resolves PDS-1393.

## Test plan

Tested locally, but minimally, with corresponding cli changes.

Compared Dockerfiles generated on main to Dockerfiles generated on this branch (specifically for Python fixtures).


pkg/build/python.go
```diff
@@ -96,35 +204,18 @@ func python(
 			&& apt-get autoremove -y && apt-get clean -y && rm -rf /var/lib/apt/lists/*
 
 		WORKDIR /airplane
-		RUN pip install "airplanesdk>=0.3.0,<0.4.0"
-		RUN mkdir -p .airplane && {{.InlineShim}} > .airplane/shim.py
+		ENV PIP_CONFIG_FILE=pip.conf
```
calufornia, line 207: IIUC moving this to the top of the Dockerfile is safe because:

1. If the user is installing requirements, they'll need this
2. If the user is not installing requirements, the presence of this env var doesn't affect anything

At the same time, it might be nice to add a 5th instruction type `Env` this is used for scenarios like this. Not necessary in this PR, but something to think about if env vars end up mattering more for node, etc.

pkg/build/python.go
```diff
@@ -96,35 +204,18 @@ func python(
...
-		{{else if .PreInstallPath}}
-		COPY {{ .PreInstallPath }} airplane_preinstall.sh
-		RUN chmod +x airplane_preinstall.sh && ./airplane_preinstall.sh
-		{{end}}
-
-		{{if .HasRequirements}}
-		COPY requirements.txt .
-		{{range .EmbeddedRequirements}}
-		COPY {{.}} .
+		{{range .InstallInstructions}}
```
calufornia, line 211: I notice that this logic is repeated in all the Dockerfiles (and I'm guessing it will be repeated for node tasks). It might be nice to create some `instructionTemplate` string that can be used in all the heredocs.

calufornia (approved):

PR #514, opened 2023-01-21
Dedupe bundle tasks

leeweisberger:
## Description
A decent portion of panther's deploy latency comes from the fact that we are including some of their tasks in multiple bundles. This causes us to have to update many more tasks at the end of the deployment.

The bundling logic creates a different bundle if two tasks have the same environment variable key with a different value. This is because the environment variable may be used at build time (we aren't sure) so we need to build them separately. **This can only happen with non-inline tasks. Inline tasks can only specify build-time env vars in the airplane.yaml which is unique per bundle**.

So this is the case we get into (and which I unit tested and fixed)
taskA -> env `foo: bar`
taskB -> env `foo:another`
taskC -> no env vars.

Previously, we were creating two bundles. Both have taskC because it doesn't conflict with either of them:
taskA, taskC
taskB, taskC

Now, we are deduping after the fact. If taskC is in two bundles with the same base/type/version (e.g. python 3.10 full), we ensure it is only included in one bundle
taskA, taskC
taskB

## Test plan
I added a fixture that replicates this issue, wrote a failing test, then fixed it.

I then deployed panthers repo, both locally and to stage.

Stage produced 393 task revisions. Local produced 79!

Here's an example of what was fixed. Before, file `operations/add_remove_customer_to_from_group/add_remove_customer_to_from_group.py` was in a bundle with a whole bunch of other things. This task has an env var that clashes with some other tasks. We were including it... plus duping all of the tasks that didn't clash in its bundle!

Now, we build it in its own bundle. The other tasks are deduped into some other bundle.

Local
image
https://ui.honeycomb.io/airplane/datasets/dev/result/qyqaQVsHTs2

Stage
image
https://ui.honeycomb.io/airplane/datasets/stage/result/9qWqkp28q1a


pkg/deploy/bundlediscover/discover.go
```diff
@@ -69,11 +69,36 @@ func (d *Discoverer) Discover(ctx context.Context, paths ...string) ([]Bundle, e
...
+					b1Targets[target] = struct{}{}
+				}
+				var b2Targets []string
+				for _, target := range b2.TargetPaths {
+					if _, ok := b1Targets[target]; !ok {
+						b2Targets = append(b2Targets, target)
+					}
+				}
+				b2.TargetPaths = b2Targets
+				dedupedBundles[j] = b2
```
ameyakhare, line 97: can there be a case where one of the bundles has no files and can be removed? no right?

leeweisberger: A bundle has to have at least one target path or else it wouldn't be a bundle. And if it didn't somehow... we'd want it to be removed

ameyakhare (approved):

pkg/deploy/bundlediscover/discover.go
leeweisberger: Kind of misunderstood. Answer is still no because the only reason multiple bundles were produced is because there are two targets that conflict which will never exist in both bundles!

PR #515, opened 2023-01-21
Views read env vars from airplane.yaml

leeweisberger:
## Description
Read views env vars from airplane.yaml config file. 

You can specify:
```
view:
  envVars:
    foo: bar
```
And we will send env var `foo: bar` to the server attached to the view.

The next step in this project is also reading env vars from the view inline config. This requires a change to the js sdk. In this PR, I'm supporting reading env vars from the inline config as soon as we change the sdk!

## Test plan
1) Added unit tests
2) I manually tested this. I was able to specify an env var in an `airplane.yaml` and send it up to be persisted in the view.


pkg/deploy/config/config.go
```diff
@@ -37,9 +37,14 @@ type PythonConfig struct {
...
 }
 
+type ViewConfig struct {
+	EnvVars TaskEnv `yaml:"envVars,omitempty" json:"envVars,omitempty"`
+}
+
 type AirplaneConfig struct {
 	Javascript JavaScriptConfig `yaml:"javascript,omitempty" json:"javascript,omitempty"`
 	Python     PythonConfig     `yaml:"python,omitempty" json:"python,omitempty"`
+	View       ViewConfig       `yaml:"view,omitempty" json:"view,omitempty"`
```
ameyakhare, line 47: do we support different js versions for views? if so, maybe we should nest this under the js config?

ameyakhare (approved):
should make sure this works in studio as well

pkg/deploy/config/config.go
leeweisberger: What I'm thinking is that `Javascript` is not the language, but the task type. It's a way to configure your javascript tasks. While view is a way to configure your view. I think this makes sense and is less confusing?

PR #518, opened 2023-01-23
Add application_name to postgres DSN

francescaleung:
## Description

Affects Postgres resources going forward; this will not affect current Postgres resources.

Resolves DTS-164.

## Test plan

Tested locally.


joshma:
> Affects Postgres resources going forward; this will not affect current Postgres resources.

👀  will we have to ask all customers to recreate their resources somehow? is there a reason why we can't do this in the builtin instead?

pkg/resources/kinds/postgres.go
```diff
@@ -157,6 +157,7 @@ func (r *PostgresResource) UpdateBaseResource(br resources.BaseResource) error {
 func (r PostgresResource) dsn() string {
 	q := url.Values{}
 	q.Set("sslmode", r.SSLMode)
+	q.Set("application_name", "Airplane")
```
calufornia, line 160: Should this be lowercase?

calufornia (approved):

pkg/resources/kinds/postgres.go
francescaleung: we uppercase it in Snowflake, I think uppercase is reasonable

francescaleung:
> > Affects Postgres resources going forward; this will not affect current Postgres resources.
> 
> eyes will we have to ask all customers to recreate their resources somehow? is there a reason why we can't do this in the builtin instead?

This is solved by https://github.com/airplanedev/airport/pull/3978.

PR #519, opened 2023-01-24
Update airplane ignore to handle yarn berry

leeweisberger:
## Description
Our patterns in our airplaneignore were incorrect for yarn berry. Docker ignore patterns are weird and what we had wasn't working. 

I updated so it works now.

## Test plan
I cloned a repo with yarn berry and confirmed that I could not deploy b/c the .yarn folder was not getting uploaded. After this change, I am able to deploy and the `releases` and `plugins` subfolders are uploaded. Other subfolders are not uploaded.

The test is commented back in in https://github.com/airplanedev/lib/pull/520


pkg/build/ignore/ignore.go
```diff
@@ -78,11 +78,11 @@ func Patterns(path string) ([]string, error) {
 		".terraform",
 		".venv",
 		".vercel",
-		".yarn",
-		"!.yarn/patches",
-		"!.yarn/plugins",
-		"!.yarn/releases",
-		"!.yarn/versions",
+		"/.yarn",
```
colinking, line 81: Why do you need the `/`? Is it because of the leading `.`?

colinking (approved):

pkg/build/ignore/ignore.go
leeweisberger: No it's not that.

We have some logic that converts this -> docker ignore format.

We were previously producing `**.yarn`. This wasn't working with the excludes patterns because you can't use ** in an excludes pattern. This was mostly just be manually testing things to figure out what works in docker.

This new change converts to -> `.yarn` instead. This will only find directories at the top level, but this is the only place you can have a .yarn file. And it works with the exclude patterns

PR #521, opened 2023-01-24
Validate schedule and resource slugs

leeweisberger:
## Description
Validate that schedule and resource slugs match the pattern.

Currently, there is no validation done as additional properties are allowed.

## Test plan
I wrote a test for a valid and an invalid schedule


pkg/deploy/discover/discover_test.go
```diff
@@ -671,6 +671,55 @@ func TestDiscover(t *testing.T) {
 			},
 			defnFilePaths: []string{fixturesPath + "/taskInline/codeOnlyTask.airplane.ts"},
 		},
+		{
+			name:  "node code definition with schedule",
+			paths: []string{"./fixtures/taskInlineSchedule/codeOnlyTask.airplane.ts"},
+			existingTasks: map[string]api.Task{
+				"collatz": {ID: "tsk123", Slug: "collatz", Kind: build.TaskKindPython, InterpolationMode: "jst"},
```
francescaleung, line 678: shouldn't these be TaskKindNode

francescaleung (approved):

PR #522, opened 2023-01-24
Create postcss file if tailwind config is present

geodn:
## Description
If we find a tailwind config, we can support tailwind by adding a postcss file. The postcss file points to the tailwind config, which ends up copied to src. We will require the tailwind config to specify content.relative = true.
## Test plan
In airport:
`go get github.com/airplanedev/lib@ecf580f391707a26e42f8c8ea52bceb2848aaa88`
`docker-compose build bundler`
`docker-compose restart builder`
In a sample view with tailwind:
`airplane --host api.airplane.so:5000 deploy` and verify that the view has tailwind styles applied


pkg/build/view.go
```diff
@@ -193,6 +193,16 @@ func viewBundle(root string, buildContext BuildContext, options KindOptions, fil
...
+	postcssConfigStr, err := PostcssConfigString("src/tailwind.config.js")
+	if err != nil {
+		return "", err
+	}
+
+	hasTailwind := false
+	tailwindPath := filepath.Join(root, "tailwind.config.js")
+	if fsx.Exists(tailwindPath) {
+		hasTailwind = true
+	}
```
leeweisberger, line 205:
```suggestion
	tailwindPath := filepath.Join(root, "tailwind.config.js")
	hasTailwind := fsx.Exists(tailwindPath)
```

pkg/build/view.go
```diff
@@ -350,6 +364,9 @@ func viewBundle(root string, buildContext BuildContext, options KindOptions, fil
 		# Copy and install dependencies.
 		COPY package*.json yarn.* /airplane/
 		RUN {{.InlinePackageJSON}} > /airplane/package.json && {{.InstallCommand}}
+		{{if .HasTailwind}}
+		RUN {{.InlinePostcssConfig}} > /airplane/postcss.config.js
```
leeweisberger, line 368: This is going to get overriden if they already have a postcss.config.js defined. We probably want to do this after we copy their files? Or we could document and tell them if they include their own postcss, it has to point at the correct config location. Thoughts? Maybe the latter?

leeweisberger (approved):

pkg/build/view.go
geodn: postcss.config.js (if created) ends up in src. Our strategy will be to tell users not to make their own postcss.config.js for now

PR #527, opened 2023-01-26
Fix python shim to catch all errors

ericliuche:
## Description
Moves the exception handling all the way outside so we can handle errors more cleanly. The motivation here was a customer's code was crashing in 
```
    mod = util.module_from_spec(spec)
    spec.loader.exec_module(mod)
```
which we weren't catching before

## Test plan
Ran locally


pkg/build/python-shim.py
```diff
@@ -32,53 +32,57 @@ def run(args):
...
+            airplane.set_output(ret)
+        except NameError:
+            # airplanesdk is not installed - gracefully print to stdout instead.
+            # This makes it easier to use the shim in a dev environment. We ensure airplanesdk
+            # is installed in production images.
+            sys.stdout.flush()
+            print("The airplanesdk package must be installed to set return values as task output.", file=sys.stderr)
+            print("Printing return values to stdout instead.", file=sys.stderr)
+            sys.stderr.flush()
+            print(json.dumps(ret, indent=2))
```
ericliuche, line 76: none of this changed, it was just dedented

pkg/build/python-shim.py
```diff
@@ -32,53 +32,57 @@ def run(args):
...
+    try:
+        run(sys.argv)
     except Exception as e:
         print(traceback.format_exc(), file=sys.stderr)
-        airplane.set_output(str(e), "error")
+        try:
+            airplane.set_output(str(e), "error")
+        except NameError:
+            # airplanesdk is not installed so we can't set the output.
+            pass
```
ericliuche, line 87: this is the only change

pkg/build/universal-python-shim.py
```diff
@@ -27,60 +28,74 @@ def run(args):
...
+            # This makes it easier to use the shim in a dev environment. We ensure airplanesdk
+            # is installed in production images.
+            sys.stdout.flush()
+            print(
+                "The airplanesdk package must be installed to set return values as task output.",
+                file=sys.stderr,
+            )
+            print("Printing return values to stdout instead.", file=sys.stderr)
+            sys.stderr.flush()
+            print(json.dumps(ret, indent=2))
```
ericliuche, line 88: none of this changed, it was just dedented

pkg/build/universal-python-shim.py
```diff
@@ -27,60 +28,74 @@ def run(args):
...
+        run(sys.argv)
     except Exception as e:
         print(traceback.format_exc(), file=sys.stderr)
-        airplane.set_output(str(e), "error")
+        try:
+            airplane.set_output(str(e), "error")
+        except NameError:
+            # airplanesdk is not installed so we can't set the output.
+            pass
         sys.exit(1)
```
ericliuche, line 101: this is the only change in this file

ameyakhare (approved):

PR #530, opened 2023-01-30
Upgrade esbuild

leeweisberger:
## Description
Upgrade esbuild to the latest version. We are really far behind and should really be up to the latest as tons of new typescript features are being supported in esbuild.

I read all of the minor (breaking) upgrades in jsbuild and don't see any issues with this upgrade. All of our tests pass. That being said I can't be 100% sure this isn't going to break anybody :/

I did add our build dependencies to a package.json (which we read in at runtime), so we should be up to date from now on w/renovate.

## Test plan
All of our unit tests pass. We have a pretty good unit test suite so this makes me reasonably comfortable this is going to be a success.

FIXES PDS-1259


pkg/build/package.json
```diff
@@ -0,0 +1,8 @@
+{
+    "dependencies": {
```
colinking, line 2: Clever

colinking (approved):

PR #531, opened 2023-01-31
Pass query params to View

stepzhou:
## Description

Get the query params and pass it to the exported View.

Resolves PDS-1470

## Test plan

* Build CLI locally using local copy of lib
* Init an inline View
* Add a console.log on the params prop
* airplane dev with built CLI
* Verify that the query params are printed

image


stepzhou:
cc: @leeweisberger

pkg/build/views/main.tsx
```diff
@@ -79,12 +80,17 @@ for (const [key, value] of getAllQueryParams()) {
   }
 }
 
+const AppWrapper = () => {
+  const { params } = useRouter();
+  return <App params={params} />;
```
leeweisberger, line 85: Why is this `params={params}` Instead of spreading the params? Not saying it's wrong, but why'd you choose that?

stepzhou: Future-proofing in case we want to add more magic props to Views in the future. Keeps it more extensible. I used the same reasoning as why for APIs it's usually better to return a map with a key even if you're just returning one thing

leeweisberger (approved):

rose-huang (approved):

PR #534, opened 2023-02-02
Support emitting metadata decorator

leeweisberger:
## Description

Support tsconfig property `emitDecoratorMetadata`. This is needed by some typescript packages like typeORM. One of our customers has asked for this.

esbuild does not support this out of the box because it doesn't support or run tsc. Context: https://github.com/evanw/esbuild/issues/257. The maintainer recommends using a plugin.

I read through the source code of the plugin we're using (https://github.com/thomaschaaf/esbuild-plugin-tsc). It only runs the "expensive" step to generate metadata if:
- There is a tsconfig.json file
- It has `emitDecoratorMetadata` set to true
So most of the time it will be a no-op :)

## Test plan
I wrote a unit test that fails before this change because it relies on introspecting generated metadata. it passes after this change and after setting `emitDecoratorMetadata` to true in the `tsconfig`


pkg/build/package.json
```diff
@@ -1,8 +1,10 @@
 {
     "dependencies": {
-        "esbuild": "0.17.5",
-        "jsdom": "21.1.0",
+        "@airplane/workflow-runtime": "0.2.33",
         "airplane": "0.2.33",
-        "@airplane/workflow-runtime": "0.2.33"
+        "esbuild": "0.17.5",
+        "esbuild-plugin-tsc": "0.3.1",
+        "jsdom": "^21.1.0",
```
colinking, line 7: Remove the ^ so this is pinned.

colinking (approved):

pkg/build/package.json
leeweisberger: @colinking I specifically didn't pin this one. My thinking is that we use `jsdom` for its super basic features and it's a very mature library that is updated pretty often. I don't really care if we're on the latest version or even if we keep up to date. The drawback of pinning is that we update more frequently which breaks the cache and causes us to have to redownload dependencies. Make sense to you?

leeweisberger: I can also leave it pinned but change the schedule to updating like once a month max?

colinking: Was just about to propose that :) SGTM!

PR #536, opened 2023-02-02
Convert views to using esbuild.js w/ plugins

leeweisberger:
## Description
Convert views to using the same esbuild script that tasks use which includes default plugins rather than the esbuild CLI.

A user was running into an issue where their view wouldn't deploy when importing a 3p css library. I fixed this earlier in our local discovery/tasks code, but did not fix it in our views code. Switch views to using esbuild which already contains a plugin that removes css files during discovery.

## Test plan
I wrote a unit test that reproduced the issue, and now it works.

---
Note that I'm going to be making a few more improvements to views build. Not trying to make them all at once


pkg/build/package.json
```diff
@@ -1,8 +1,13 @@
 {
     "dependencies": {
+        "@airplane/views": "1.14.3",
```
leeweisberger, line 3: Store views dependencies here so that they can be handled by renovate. We're pinning the dependencies that we hope to control and ^ dependencies that we care less about having exact versions of

pkg/build/view.go
```diff
@@ -218,33 +218,19 @@ func viewBundle(root string, buildContext BuildContext, options KindOptions, fil
...
-		}
-		if devDepsOk {
-			if _, ok := devDeps[pkg]; ok {
-				hasPackage = true
-			}
-		}
-		if !hasPackage {
-			packagesToAdd = append(packagesToAdd, pkg)
-		}
+	packagesToAdd, err := getRequiredViewsDependencies([]string{packageJSONPath})
```
leeweisberger, line 221: This is just a refactor which pulls the versions from the package.json and puts the code in a helper function

pkg/build/view.go
```diff
@@ -399,6 +393,36 @@ func viewBundle(root string, buildContext BuildContext, options KindOptions, fil
...
+
+	requiredDepsMap := map[string]string{
+		"@airplane/views":      buildToolsPackageJSON.Dependencies["@airplane/views"],
+		"react":                buildToolsPackageJSON.Dependencies["react"],
+		"react-dom":            buildToolsPackageJSON.Dependencies["react-dom"],
+		"vite":                 buildToolsPackageJSON.Dependencies["vite"],
+		"esbuild":              buildToolsPackageJSON.Dependencies["esbuild"],
+		"typescript":           buildToolsPackageJSON.Dependencies["typescript"],
+		"@vitejs/plugin-react": buildToolsPackageJSON.Dependencies["@vitejs/plugin-react"],
+	}
```
ameyakhare, line 416: maybe refactor this into a list and range?

renovate.json
```diff
@@ -12,12 +12,8 @@
   ],
   "packageRules": [
     {
-      "matchPackageNames": ["jsdom"],
-      "schedule": ["before 5am on the first Thursday of the month"]
-    },
-    {
```
ameyakhare, line 18: why was this removed?

ameyakhare (approved):

renovate.json
leeweisberger: It was invalid. Unrelated. Just looping it in

PR #537, opened 2023-02-02
Add http code-specific errors

calufornia:
## Description

Adds error structs that we use in our prod API to lib, so that we can also use them in the local dev server in CLI. A follow-up change will be made to swap our prod API to also use this package.

## Test plan
n/a


pkg/errorsx/unwrap.go
```diff
@@ -0,0 +1,43 @@
...
+// The former interface is from Go's standard library errors package.
+// The latter interface is from `github.com/pkg/errors`.
+//
+// If the error does not implement either interface, the original error will
+// be returned. If the error is nil, nil will be returned without further
+// investigation.
+//
+// Unlike `pkg/errors.Cause`, this supports both interfaces.
+// This method traverses the error recursively rather than unwrapping at most once.
+func Cause(err error) error {
```
colinking, line 25: IMO, we should stop using `Cause` and switch to `errors.Is/As`. Switching API is going to take some work, but can we at least avoid adding it to `lib`?

pkg/errorsx/http.go
```diff
@@ -0,0 +1,81 @@
+package errorsx
+
+type Code string
```
colinking, line 3: We already have a similar struct in `lib` here: https://github.com/airplanedev/lib/blob/76f6650c7c7771f989e8ebbe0d688f876259f876/pkg/api/http/err.go#L12

What do you think of using that? 

We could then create helpers for returning errors with the status code pre-filled:

```
var err http.ErrStatusCode = http.NewErrBadRequest("", ...)
```

colinking (changes requested):

pkg/errorsx/http.go
calufornia: Ah, didn't know we had this! I'll use this instead

calufornia: I decided that the pre-filled error might be unnecessary—I'm thinking that we just create the helper functions and then the caller can do something like `NewErrBadRequest("malformed input")`, LMK what you think

pkg/api/http/err.go
```diff
@@ -53,3 +51,51 @@ func NewErrStatusCodeFromResponse(resp *http.Response) error {
 
 	return errsc
 }
+
+func NewErrStatusCode(statusCode int, msg string, errorCode ...string) ErrStatusCode {
```
colinking, line 55: The intention behind the variadic args was for making it easier to format the message!

```suggestion
func NewErrStatusCode(statusCode int, msg string, args ...string) ErrStatusCode {
```

pkg/api/http/err.go
```diff
@@ -53,3 +51,51 @@ func NewErrStatusCodeFromResponse(resp *http.Response) error {
...
+	var ec string
+	if len(errorCode) > 0 {
+		ec = errorCode[0]
+	}
+	return ErrStatusCode{
+		StatusCode: statusCode,
+		Msg:        msg,
```
colinking, line 62:
```suggestion
  if len(args) > 0 {
  	msg = fmt.Sprintf(msg, ...args)
  }
	return ErrStatusCode{
		StatusCode: statusCode,
		Msg:        msg,
```

pkg/api/http/err.go
colinking, line 55: Let's not export this -- folks should use one of the functions below instead.

pkg/api/http/err.go
```diff
@@ -53,3 +51,51 @@ func NewErrStatusCodeFromResponse(resp *http.Response) error {
...
+		ErrorCode:  ec,
```
colinking, line 63: Setting an error_code is very rare, so I'm fine if it's more work. Seems reasonable that the user would just do `err := /*..*/; err.ErrorCode = "..."`

```suggestion
```

pkg/api/http/err.go
```diff
@@ -24,9 +24,7 @@ func (e ErrStatusCode) Error() string {
 }
 
 func NewErrStatusCodeFromResponse(resp *http.Response) error {
-	errsc := ErrStatusCode{
-		StatusCode: resp.StatusCode,
-	}
+	errsc := NewErrStatusCode(resp.StatusCode, "", "")
```
colinking, line 27: (tbh I think this is less clear)

colinking (approved):

PR #543, opened 2023-02-06
Fix JS task execution on Windows

calufornia:
## Description

JS tasks do not work in Windows—we get an error saying that the shim failed to build. This is because we attempt to manually "stringify" the entrypoints argument of the esbuild script by wrapping its elements in quotes. This is error-prone and does not work on Windows, because backslashes need to be properly escaped.

Instead, we just construct the underlying object and just use on `json.Marshal` to properly escape backslashes, etc. This maintains a better contract between `esbuild.js` and callers of it. cc @leeweisberger 

## Test plan
Tested on Windows. Verified that changes do not affect POSIX.


pkg/build/esbuild.js
```diff
@@ -37,7 +39,7 @@ const removeCSS = {
 
 const entryPoints = JSON.parse(process.argv[2]);
 const target = process.argv[3];
-const external = JSON.parse(process.argv[4]);
+const external = JSON.parse(process.argv[4]) || [];
```
ericliuche, line 42: what's happening here?

ericliuche (approved):
How easy is it to add a test to run on windows? Fine to punt for now, but lets at least make a task for it

pkg/build/esbuild.js
calufornia: This is the edge case when there's no external dependencies—`JSON.Marshal` marshals empty slices to null, and so it also gets parsed as null. But we require `external` being iterable below (line 65), so we default to an empty array

calufornia:
> How easy is it to add a test to run on windows? Fine to punt for now, but lets at least make a task for it

It shouldn't be too bad if we have testing on other operating systems in place, but yep I'll make a task to track it

calufornia:
Need to make this work with deploys, which also use `esbuild.js`

pkg/build/node.go
```diff
@@ -1199,7 +1205,7 @@ func nodeBundle(
 			node /airplane/.airplane/esbuild.js \
 			'["/airplane/.airplane/universal-shim.js"]' \
```
calufornia, line 1206: We can _also_ marshal this argument, but I'm inclined to leave it as is since it's static and we'd need to add another parameter to the template config

pkg/build/node.go
```diff
@@ -1199,7 +1205,7 @@ func nodeBundle(
...
 			node{{.NodeVersion}} \
-			'[{{.ExternalFlags}}]' \
+			'{{.External}}' \
```
calufornia, line 1208: We still need to wrap `{{.External}}` in single quotes or else the shell will interpret the value as an array (`text/template` does not do further wrapping of string values)

calufornia:
I got this working for deploys as well—in hindsight, I don't think this was strictly necessary (since the bundler is never running on a windows machine), but it's probably better to have both dev + build aligned.

leeweisberger (approved):
Oh man good catch.

I wonder if we can run our unit tests on windows (which I think would have caught this?)

leeweisberger:
Did you confirm that you can build a view with this manually? the views tests aren't as strict as the task tests

PR #547, opened 2023-02-08
Make builtins client download logic thread-safe

calufornia:
## Description

Exports the `Download` method of the builtin client so that it can be called by the executor when attempting to refresh the builtins binary. Also adds a mutex so that multiple callers cannot concurrently download the binary and override each other, which may help with errant `signal: killed` errors.

## Test plan
Tested in conjunction with CLI changes in https://github.com/airplanedev/cli/pull/951


kellymtrinh (approved):
As discussed! 🥳

PR #549, opened 2023-02-09
Add method for editing inline task config

colinking:
## Description

This PR extends the `runtime` interface to support editing inline task config. It currently only supports editing JavaScript and TypeScript inline config. It does not yet support Python, Views, nor YAML files. Those will be added in future PRs. I'll also be following up with more test cases.

See: DTS-286

## Test plan

See unit tests. This can also be tested via this CLI command: https://github.com/airplanedev/cli/pull/967


colinking:
* main
  * **PR #549** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/549" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/549?utm_source=stack-comment).

pkg/runtime/javascript/javascript_test.go
```diff
@@ -135,3 +141,197 @@ func TestVersion(t *testing.T) {
...
+		// TODO: confirm various error cases return user-friendly errors
+		// TODO: ignore non-airplane call expressions even with matching names
+		// TODO: detect if fields are explicitly set to a default value + retain the value if so
+		// TODO: test case where we can't edit the task for some reason (e.g. parsing), and that we get a sensible error back
+		// TODO: support serializing parameters as type-only strings
+		// TODO: support views
+		// TODO: multiple fields with the same field name
+		// TODO: audit all possible errors
+		// TODO: audit unexpected config format error
+		// TODO: string literal with double/single quotes in it
```
colinking, line 302: I'll follow-up with these test cases in the next PR.

pkg/runtime/javascript/transformer/transformer.ts
```diff
@@ -0,0 +1,387 @@
...
+        ]);
+      } else {
+        propertyValue = builders.stringLiteral(envVarValue["value"] ?? "");
+      }
+      value.properties.push(buildObjectProperty(envVar, propertyValue));
+    }
+    output.properties.push(buildObjectProperty("envVars", value));
+  }
+
+  const defaultTimeout = def.runtime === "workflow" ? 0 : 3600;
```
ameyakhare, line 174: is it possible to remove this business logic and encompass it in the def that's being passed down to this serializer? I'm worried we may change these values and it won't be reflected here

pkg/runtime/javascript/transformer/transformer.ts
```diff
@@ -0,0 +1,387 @@
...
+            buildObjectProperty("default", builders.booleanLiteral(p.default))
+          );
+        }
+        return buildObjectProperty(p.slug, parameter);
+      })
+    );
+    output.properties.push(buildObjectProperty("parameters", parameters));
+  }
+
+  if (def.runtime && def.runtime !== "standard") {
```
ameyakhare, line 132: ditto on comment below

ameyakhare (approved):
overall looks fine, we should just be wary of the limitations of inline-discovered things being written back as literal values

.gitignore
```diff
@@ -22,9 +22,6 @@ __pycache__/
 # IntelliJ editors
 .idea/
 
-# VSCode editors
```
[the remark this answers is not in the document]
colinking: Did this so we can commit the `.vscode` folder which includes some recommended config.

pkg/runtime/javascript/transformer/transformer.ts
```diff
@@ -0,0 +1,387 @@
...
+): namedTypes.ObjectExpression => {
+  const output = builders.objectExpression([]);
+
+  {
+    const value = builders.stringLiteral(def.slug);
+    value.comments = getComments(input, "slug");
+    output.properties.push(buildObjectProperty("slug", value));
+  }
+
+  if (def.name) {
```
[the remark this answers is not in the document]
colinking: The native JS would get replaced, yeah. It gets complicated quickly if we wanted to support that and it'll never be perfect. For now, I want to avoid that and see if it becomes a bigger issue later.

pkg/runtime/javascript/transformer/transformer.ts
colinking: I'm not too concerned with this changing, since "standard" is by definition the default value.

pkg/runtime/javascript/transformer/transformer.ts
colinking: Possibly (maybe a second arg with "default values")? For now, since it's just the timeout, I'm going to leave as-is. I had thought I would need to push down more default values, but it ended up just being timeout.

PR #550, opened 2023-02-09
Always include bundle dependencies in shim package.json

calufornia:
## Description

If a user has the `typescript` dependency in their `package.json` (or any of the other dependencies below), we previously would not include it in our shim package.json. This was problematic because for bundle builds, we require `typescript`. We make a change to always include these dependencies in the shim package.json so that our esbuild script can run properly.

## Test plan
Tested with a local bundler and verified that inline view deploys succeeded.


pkg/build/node.go
```diff
@@ -693,6 +687,15 @@ func GenShimPackageJSON(rootDir string, packageJSONs []string, isWorkflow, isBun
 		delete(pjson.Dependencies, dep)
 	}
 
+	// These dependencies must be included in order to build the universal task shim. We do not install user
+	// dependencies before building the shim in bundle builds, so we include them here.
+	if isBundle {
```
yolken-airplane, line 692: Will there be any downstream problems if users have a different version of one of these dependencies in their package.json files?

yolken-airplane: i.e., should we use their version instead of ours?

calufornia: good point, let me do that

yolken-airplane (approved):

PR #552, opened 2023-02-10
Support compressing request bodies

yolken-airplane:
## Description
This change updates our shared http client to compress large request bodies. Once this is in, we can update the lib version in airport and use this to compress the log and output appending requests made by the runbroker.

## Test plan
Testing completed successfully via running locally in https://github.com/airplanedev/airport/pull/4064.


pkg/api/http/client.go
```diff
@@ -321,8 +338,28 @@ func (c Client) do(ctx context.Context, method string, url string, req []byte, o
 	// that encounter certain connection errors:
 	// https://github.com/golang/go/blob/f721fa3be9bb52524f97b409606f9423437535e8/src/net/http/request.go#L1441-L1455
 	if req != nil {
+		var serializedBody []byte
```
colinking, line 341: Do this before we create `NewRequestWithContext` and update the value of `req` instead so that we don't use `req` in one place and `serializedBody` in another.

pkg/api/http/client.go
```diff
@@ -321,8 +338,28 @@ func (c Client) do(ctx context.Context, method string, url string, req []byte, o
...
+
+		// Only compress if the caller has opted-in, the request is big enough, and
+		// there isn't already a Content-Encoding set on the request (meaning that
+		// it was already compressed upstream).
+		if opts.CompressBody &&
+			len(req) >= compressionSizeThreshold &&
+			opts.Headers["Content-Encoding"] == "" {
```
colinking, line 348: Note that you can also set headers via `Client.Headers`. Maybe just error if `Content-Encoding` is set as a header in either place instead? I don't see any reason for setting it directly.

pkg/api/http/client_test.go
```diff
@@ -124,6 +125,90 @@ func TestClientPost(t *testing.T) {
...
+		require.NoError(err)
+
+		// The server automatically decompresses the body, so it's not
+		// gzipped here.
+		require.Equal(reqBody, body)
+
+		require.Equal(req.Header.Get("X-Airplane-Client-Kind"), "test")
+		require.Equal(req.Header.Get("X-Airplane-Client-Version"), "1")
+		require.Equal(req.Header.Get("X-Req-Opts-Header"), "hello")
+		require.Equal(req.Header.Get("User-Agent"), "airplane/test/1")
```
colinking, line 149: I'd remove the unrelated assertions.

pkg/api/http/client_test.go
```diff
@@ -124,6 +125,90 @@ func TestClientPost(t *testing.T) {
...
+	}))
+	defer server.Close()
+
+	client := NewClient(ClientOpts{
+		Headers:   requiredHeaderValues,
+		UserAgent: "airplane/test/1",
+	})
+	body, err := client.Post(ctx, server.URL+"/foobar", reqBody, ReqOpts{
+		Headers: map[string]string{
+			"X-Req-Opts-Header": "hello",
```
colinking, line 163: This is unrelated.

pkg/api/http/client_test.go
```diff
@@ -124,6 +125,90 @@ func TestClientPost(t *testing.T) {
...
+		// The server automatically decompresses the body, so it's not
+		// gzipped here.
+		require.Equal(reqBody, body)
+
+		require.Equal(req.Header.Get("X-Airplane-Client-Kind"), "test")
+		require.Equal(req.Header.Get("X-Airplane-Client-Version"), "1")
+		require.Equal(req.Header.Get("X-Req-Opts-Header"), "hello")
+		require.Equal(req.Header.Get("User-Agent"), "airplane/test/1")
+		require.Equal(req.Header.Get("Content-Encoding"), "gzip")
+		require.NotEmpty(req.Header.Get("Idempotency-Key"))
```
colinking, line 151: Unrelated.

pkg/api/http/client_test.go
```diff
@@ -124,6 +125,90 @@ func TestClientPost(t *testing.T) {
...
+
+	client := NewClient(ClientOpts{
+		Headers:   requiredHeaderValues,
+		UserAgent: "airplane/test/1",
+	})
+	body, err := client.Post(ctx, server.URL+"/foobar", reqBody, ReqOpts{
+		Headers: map[string]string{
+			"X-Req-Opts-Header": "hello",
+		},
+		UserAgent:    "airplane/test/1",
```
colinking, line 165: Unrelated (it's set in client)

colinking (approved):
If you agree this should be the default for all requests, then LGTM. Otherwise, let's discuss!

pkg/api/http/client.go
```diff
@@ -137,6 +145,15 @@ type ReqOpts struct {
...
 	// "airplane/cli/v0.1.4 team/tea123".
 	UserAgent string
+
+	// Whether to gzip compress the body of the request if it's at or above
+	// compressionSizeThreshold in size. If the body is smaller then this,
+	// compression is not done.
+	//
+	// Note that this is done without first checking whether the server accepts
+	// gzip bodies. Our API does, so it's safe to use request-side compression, but
+	// third-party services might not allow this.
```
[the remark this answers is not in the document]
yolken-airplane: Yup, sounds good, will just do that.

pkg/api/http/client.go
yolken-airplane: Fixed.

pkg/api/http/client_test.go
yolken-airplane: Done.

pkg/api/http/client_test.go
yolken-airplane: Removed it.

pkg/api/http/client_test.go
yolken-airplane: Removed it.

pkg/api/http/client_test.go
yolken-airplane: Fixed.

pkg/api/http/client.go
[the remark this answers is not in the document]
yolken-airplane: Oooh, so actually, I was making the requests incorrectly. If I make the requests correctly (i.e., fixing your comment below), then the server does not automatically decompress the bodies, and we need to manually check the content-encoding and do it ourselves (see https://stackoverflow.com/questions/64176224/gzip-request-body-auto-decompress).

Given this, is it ok if I keep making compressed bodies an opt-in? Then, we can just enable it for the appendLogs and appendOutputs calls, and make sure that the server endpoints can handle decompression properly.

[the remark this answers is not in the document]
colinking: I would expect that adding a middleware to check `Content-Encoding` and decompress the request body would be a one-liner. Any reason not to do that for all endpoints? Or are you proposing we roll it out to logs/outputs first and then the remainder of our endpoints after if that went well?

pkg/api/http/client.go
```diff
@@ -312,6 +334,26 @@ func (c Client) doJSON(ctx context.Context, method string, url string, req []byt
 }
 
 func (c Client) do(ctx context.Context, method string, url string, req []byte, opts ReqOpts) ([]byte, http.Header, error) {
+	// Compress the body if opts.CompressBody is true, the request is big enough,
+	// and there isn't already a Content-Encoding set on the request (meaning that
+	// it was already compressed upstream).
+	if opts.CompressBody &&
+		len(req) >= compressionSizeThreshold &&
+		opts.Headers["Content-Encoding"] == "" {
```
colinking, line 342: IMO, let's do this + remove the TODO?

```suggestion
	contentEncoding = c.Headers["Content-Encoding"]
	if c := opts.Headers["Content-Encoding"]; c != "" {
		contentEncoding = c
	}
	// Compress the body if opts.CompressBody is true, the request is big enough,
	// and there isn't already a Content-Encoding set on the request (meaning that
	// it was already compressed upstream).
	if opts.CompressBody &&
		len(req) >= compressionSizeThreshold &&
		contentEncoding == "" {
```

pkg/api/http/client.go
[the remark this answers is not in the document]
yolken-airplane: Ahh, ok, didn't think of the middleware-based approach (was going to implement it lower-down). Let me add that on the API side, then come back to this change. Thanks.

pkg/api/http/client.go
yolken-airplane: Done.

PR #556, opened 2023-02-17
View yarn workspace

leeweisberger:
## Description
Support deploying views within yarn workspaces.

This took two steps:
1. Set the correct root directory during discovery. This logic should be the same as tasks.
2. Build the view with the same yarn workspace logic we have for tasks. This required a bit of a refactor.

In addition, I refactored the view build process a bit.

## Test plan
I added a unit test that tests that the view builds. Also add a unit test to test that discovery works.

To test that the view renders correctly, I deployed the view with a local version of lib. A regular view worked. A view with yarn workspaces worked as well.


pkg/build/node.go
```diff
@@ -85,6 +85,33 @@ func getNodeBundleBuildInstructions(
 		},
 	}
 
+	installInstructions, err := getNodeInstallInstructions(root, "/airplane")
```
leeweisberger, line 88: This was a refactor. Pulled this logic out and reusing it for taks and views

pkg/build/node_externals.go
```diff
@@ -104,10 +104,10 @@ func GetPackageCopyInstructions(baseDir string, pathPackageJSONs []string) ([]In
 	}
 
 	for srcPath := range srcPaths {
-		destPath := filepath.Join("/airplane", srcPath)
+		destPath := filepath.Join(dest, srcPath)
 		// Docker requires that the destination ends with a slash
-		if !strings.HasSuffix(destPath, "/") {
-			destPath = destPath + "/"
+		if !strings.HasSuffix(destPath, string(filepath.Separator)) {
+			destPath = destPath + string(filepath.Separator)
```
leeweisberger, line 110: no op

pkg/build/view.go
```diff
@@ -201,47 +201,34 @@ func viewBundle(root string, buildContext BuildContext, options KindOptions, fil
...
-		}
-		if err := json.Unmarshal([]byte(packageJSONFile), &packageJSON); err != nil {
-			return "", errors.Wrap(err, "parsing package JSON")
-		}
-	}
-	packageJSONMap, ok := packageJSON.(map[string]interface{})
-	if !ok {
-		packageJSON = map[string]interface{}{}
-		packageJSONMap = packageJSON.(map[string]interface{})
+	rootPackageJSONPath := filepath.Join(root, "package.json")
```
leeweisberger, line 204: I changed our strategy here. This is a decent refactor but it's better.

Instead of adding shim/build deps to the user's package.json, we create a shim package.json. 

In the dockerfile, we create structure:

/airplane
   package.json - shim with shim/build deps
   .airplane-buildtools/ - contains build tools. Can only use shim package.json
  src/
    <user_code> - Can use user's package.json or the shim package.json
    package.json - user's package.json.

pkg/build/view.go
```diff
@@ -201,47 +201,34 @@ func viewBundle(root string, buildContext BuildContext, options KindOptions, fil
...
+	packageJSONs, usesWorkspaces, err := GetPackageJSONs(rootPackageJSONPath)
+	if err != nil {
+		return "", err
 	}
 
-	packagesToAdd, err := getRequiredViewsDependencies([]string{packageJSONPath})
+	ii, err := getNodeInstallInstructions(root, "/airplane/src")
```
leeweisberger, line 210: Here's where we re-use all the logic from tasks

pkg/build/view.go
```diff
@@ -251,15 +238,17 @@ func viewBundle(root string, buildContext BuildContext, options KindOptions, fil
 	}
 	esbuildFlags := strings.Join(flags, ", ")
 
-	directoryToBuildTo := ".airplane"
+	// Build code into the src directory where user deps are installed so that
+	// it can access user deps.
+	directoryToBuildTo := "/airplane/src/.airplane"
```
leeweisberger, line 243: As the comment says, this has to be under /src so that when we run the built tasks for discovery, they can use all deps. We build to a "hidden" folder

pkg/build/views/gen_view.sh
```diff
@@ -11,6 +11,6 @@ for entrypoint in $entrypoints; do
     # Create an index.html file for each entrypoint.
     cp $indexhtml "${entrypoint}"
 
-    # Create a main.tsx file for each entrypoint. Replace {{.Entrypoint}} with /src/path/to/entrypoint
-    sed -e "s|{{.Entrypoint}}|\/src\/$entrypoint|" $maintsx > "${entrypoint}/main.tsx"
+    # Create a main.tsx file for each entrypoint. Replace {{.Entrypoint}} with /path/to/entrypoint
+    sed -e "s|{{.Entrypoint}}|\/$entrypoint|" $maintsx > "${entrypoint}/main.tsx"
```
leeweisberger, line 15: tbh not sure why I had to change this. I think it's b/c we're running vite from `/airplane/src` instead of `/airplane`.

pkg/deploy/discover/code_view_discover.go
```diff
@@ -167,7 +157,7 @@ func (dd *CodeViewDiscoverer) GetViewRoot(ctx context.Context, file string) (str
 		return "", build.BuildContext{}, err
 	}
 
-	return root, build.BuildContext{
+	return pm.RootDir, build.BuildContext{
```
leeweisberger, line 160: Use the same logic for discovery we are using for tasks! This is the same as the logic I deleted above PLUS if there is a package.json with an `airplane` field with a `root` field, we respect that

calufornia (approved):

pkg/build/view.go
```diff
@@ -358,70 +354,69 @@ func viewBundle(root string, buildContext BuildContext, options KindOptions, fil
...
-
-	// Allow users to override dependencies.
-	// TODO: don't do this. We want to install our own versions of "toolchain"
-	// dependencies (esbuild, vite, etc) to ensure they're compatible with our
-	// build system. We're doing this for now because we add these deps to the
-	// user's package.json rather than creating a higher package.json.
-	for dep := range deps {
+	// Allow users to override shim dependencies. If the user has specified a dependency, we won't
+	// install it for them and will rely on their version instead.
+	for dep := range existingDeps {
```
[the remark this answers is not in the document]
leeweisberger: These are dependencies needed for the shim which the user will probably have installed, but if they don't then we want to install for them so that the code doesn't break.

If we install two versions of this, it will just use the user's version anyways because this package.json and the user's package.json will be in scope and the user's is closer.

Even if we could use two different versions, I don't think that's a good idea. e.g. you're only supposed to have 1 version of react installed and things break if you have multiple. Same with @airplane/views - I don't really want to create a view that uses multiple versions of this dep it might get us in weird places

PR #557, opened 2023-02-22
Optimize local javascript task execution

calufornia:
## Description

Speeds up local execution of JS+TS tasks by optimizing how we build local JS tasks. Because shim dependencies and the shim itself should not change between runs (or tasks), we don't attempt to install shim dependencies or build the shim if they already exist.

However, if our build script (`esbuild.js`) or any of the build dependencies change, we do want to rebuild the shim. To detect this scenario, we track a dependency hash file `dep-hash` the detects whether or not either of these has changed since the last run, and if so we rebuild the shim. This is similar to how we force `Vite` to regenerate its dependency cache. (https://github.com/airplanedev/cli/pull/886).

Note that `npm install` does not reinstall dependencies if they're already installed, but I've found that it still requires ~1s to check. Checking the dependency hash to see if we need to reinstall is much quicker.

Also adds more debug logs to see where JS task execution might be bottle-necked. 

Also writes the built entrypoint to an individual run subdirectory instead of the task subdirectory—we were seeing some occasional race conditions where one run was causing the deletion of the task directory while another run was occurring, causing the second run to error. We now keep a task subdirectory around (similar to how we keep a view subdirectory in `.airplane-view`, and cleanup run subdirectories on the fly.

After these changes, the `.airplane/` directory looks like the following:
```
.airplane/
|_dep-hash
|_esbuild.js
|_node_modules
|_package.json
|_package-lock.json
|_my_task
   |_run1234
      |_{entrypoint}
   |_run2345
      |_{entrypoint}
```

Resolves DTS-329

## Test plan
Tested local JS execution and verified that run preparation time (after the 1st for a task across all Studio sessions) went from 3-7s to 0.4-0.8 seconds.


pkg/build/node.go
```diff
@@ -670,7 +670,7 @@ type shimPackageJSON struct {
 
 // GenShimPackageJSON generates the `package.json` that contains the dependencies required for the shim to run. If the
 // dependency is satisfied by a parent directory (i.e. the user's code), then no need to include it here.
-func GenShimPackageJSON(rootDir string, packageJSONs []string, isWorkflow, isBundle bool) ([]byte, error) {
+func GenShimPackageJSON(rootDir string, packageJSONs []string, isWorkflow, isBundle, fallbackToUserDeps bool) ([]byte, error) {
```
calufornia, line 673: @leeweisberger I noticed that checking the `package.json` files for build dependencies (e.g `getLockPackageVersion`) takes a significant amount of time (~1.5s total). I don't think we need to override build deps with user deps, and is something I think we wanted to remove regardless. I added `fallbackToUserDeps` for backwards compatibility (it's `true` everywhere except when building locally now), but if we want we can just remove that logic altogether.

pkg/build/node.go
```diff
@@ -521,7 +521,7 @@ func node(
 		return "", err
 	}
 
-	pjson, err := GenShimPackageJSON(root, packageJSONs, isWorkflow, false)
+	pjson, err := GenShimPackageJSON(root, packageJSONs, isWorkflow, false, true)
```
leeweisberger, line 524: I know it's annoying, but can we move these into a config object? Too many params to keep track of

pkg/build/node.go
leeweisberger: sgtm. Can you leave a todo to resolve this? I think I have a task to resolve it in a couple of cycles

pkg/build/node_externals.go
```diff
@@ -526,7 +526,7 @@ func getLockPackageVersion(
 		return yarnVersion
 	}
 	log.Printf(
-		"Error getting %q version from yarn: %+v\n",
+		"Error getting %q version from yarn: %v\n",
```
leeweisberger, line 529: I know you kind of removed this, but why are we printing this out? I feel like it is internal and shouldn't be logged?

pkg/runtime/javascript/javascript.go
```diff
@@ -381,26 +386,19 @@ func (r Runtime) FormatComment(s string) string {
...
+	start := time.Now()
 	checkNodeVersion(ctx, logger, opts.KindOptions)
 
 	root, err := r.Root(opts.Path)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	airplaneDir, taskDir, closer, err := airplane_directory.CreateTaskDir(root, opts.TaskSlug)
+	airplaneDir, taskDir, _, err := airplane_directory.CreateTaskDir(root, opts.TaskSlug)
```
leeweisberger, line 397: This is the part you referred to in the description where we are keeping this around?

pkg/runtime/javascript/javascript.go
```diff
@@ -414,22 +412,39 @@ func (r Runtime) PrepareRun(ctx context.Context, logger logger.Logger, opts runt
...
 	}
 	if err := os.WriteFile(filepath.Join(airplaneDir, "package.json"), pjson, 0644); err != nil {
 		return nil, nil, errors.Wrap(err, "writing shim package.json")
 	}
-	cmd := exec.CommandContext(ctx, "npm", "install")
-	cmd.Dir = airplaneDir
-	logger.Debug("Running %s (in %s)", strings.Join(cmd.Args, " "), cmd.Dir)
-	out, err := cmd.CombinedOutput()
+
+	depsChanged, err := CheckDepHash(airplaneDir)
```
leeweisberger, line 423: Can you call this like `buildDepsChanged` or something? Cause it isn't checking that user deps changed?

pkg/runtime/javascript/javascript.go
```diff
@@ -414,22 +412,39 @@ func (r Runtime) PrepareRun(ctx context.Context, logger logger.Logger, opts runt
...
 	if err != nil {
-		logger.Log(strings.TrimSpace(string(out)))
-		return nil, nil, errors.New("failed to install shim deps")
+		return nil, nil, err
 	}
 
+	// Check if shim dependencies are installed already _and_ if they are up-to-date.
+	if _, err := os.Stat(filepath.Join(airplaneDir, "node_modules")); (err == nil && !depsChanged) || errors.Is(err, os.ErrNotExist) {
```
leeweisberger, line 429: why is this confusing me: You're saying if there is no node_modules reinstall (makes sense) OR if the deps haven't changed reinstall (doesn't make sense). Should it be if the deps have changed reinstall?

pkg/runtime/javascript/javascript.go
```diff
@@ -446,55 +461,77 @@ func (r Runtime) PrepareRun(ctx context.Context, logger logger.Logger, opts runt
...
-		fmt.Sprintf(`[%s]`, strings.Join(external, ", ")),
-		builtShimPath,
-	)
-	cmd.Dir = airplaneDir
-	logger.Debug("Running %s (in %s)", strings.Join(cmd.Args, " "), cmd.Dir)
-	out, err = cmd.CombinedOutput()
-	if err != nil {
-		logger.Log(strings.TrimSpace(string(out)))
-		return nil, nil, errors.New("failed to build task shim")
+	if _, err := os.Stat(builtShimPath); (err == nil && !depsChanged) || errors.Is(err, os.ErrNotExist) {
```
leeweisberger, line 472: same here

leeweisberger (changes requested):
Good stuff.

Just blocking on the logic for building / using cached versions of the build deps/shim

pkg/build/node_externals.go
calufornia: Hmm yeah good point. Are you suggesting we remove the logs entirely? We don't use log.Printf anywhere else in lib. I suppose the right thing to do would be to thread down a logger

pkg/runtime/javascript/javascript.go
calufornia: Correct— I think it's OK to leave for now (since it's inside a hidden folder, and I don't expect users to have too many tasks)

pkg/runtime/javascript/javascript.go
calufornia: Sorry, `depsChanged` was incorrectly named—`checkDepHash` returns whether or not the hashes are _equal_, not different. Updated the variable name to reflect this

pkg/build/node_externals.go
leeweisberger: Yeah remove the log entirely

leeweisberger: Yeah remove the log entirely

leeweisberger (approved):

PR #560, opened 2023-02-22
Pass views token in additional env vars

kellymtrinh:
## Description
Get the views token from the iframe query params and pass to setEnvVars.
## Test plan
- [x] Verified that env vars are not affected.
- [x] Tested with web change that passes views token in, and verified that when the token is provided, the view can execute a restricted task. Without the token passed in, the task fails with an error about execute rules.


pkg/build/views/main.tsx
```diff
@@ -53,6 +53,7 @@ setEnvVars(
   {
     AIRPLANE_TUNNEL_TOKEN: import.meta.env.AIRPLANE_TUNNEL_TOKEN,
     AIRPLANE_SANDBOX_TOKEN: getQueryParam("__airplane_sandbox_token"),
+    AIRPLANE_VIEW_TOKEN: getQueryParam("__airplane_view_token"),
```
leeweisberger, line 56: Go ahead and filter this param out of the env vars in the code block below. Please do the same with the sandbox token which I think we forgot to do.

leeweisberger: Also how are we going to do this in studio?

leeweisberger (changes requested):

pkg/build/views/main.tsx
kellymtrinh: Restricted tasks aren't enforced for local tasks in studio (we'll just show a message saying once deployed it's restricted). Otherwise remote calls will be coming from the studio via the same airplane token it currently uses to call the API

PR #563, opened 2023-02-23
Allow optional env vars during discovery

ameyakhare:
Part of [DTS-460](https://linear.app/airplane/issue/DTS-460/environment-variables-dont-work-in-inline-config-in-studio)

## Description
Allows the passing of env vars during discovery. The CLI will use this to pass env vars to the parser during discovery for local dev.

## Test plan
I tested this end to end on a parameter variable with a dynamic default value and it works.


pkg/deploy/discover/build.go
```diff
@@ -95,7 +95,7 @@ type ParsedJSConfigs struct {
 }
 
 // Extracts view and code configs from a compiled JS file.
-func extractJSConfigs(file string) (ParsedJSConfigs, error) {
+func extractJSConfigs(file string, env []string) (ParsedJSConfigs, error) {
```
joshma, line 98: are we going to pass the full set of env vars here? i guess it's whatever subset is relevant / possible?

pkg/deploy/discover/code_view_discover.go
```diff
@@ -56,7 +56,7 @@ func (dd *CodeViewDiscoverer) GetViewConfig(ctx context.Context, file string) (*
 		return nil, err
 	}
 
-	parsedConfigs, err := extractJSConfigs(compiledJSPath)
+	parsedConfigs, err := extractJSConfigs(compiledJSPath, nil)
```
calufornia, line 59: How come this is nil for views? I feel like we should technically support the same functionality for inline views?

calufornia: If this is just to support tasks for now, can you at least add a TODO?

ameyakhare: added the ability to pipe in env vars during discovery for views as well

pkg/deploy/discover/build.go
ameyakhare: only whatever is relevant/is documented as available during discovery in studio (which will be done in the CLI)

pkg/deploy/discover/build.go
```diff
@@ -107,7 +107,9 @@ func extractJSConfigs(file string) (ParsedJSConfigs, error) {
 	}
 
 	// Run parser on the file
-	out, err := exec.Command("node", tempFile.Name(), file).Output()
+	parserCmd := exec.Command("node", tempFile.Name(), file)
+	parserCmd.Env = env
```
calufornia, line 111: Just double checking, but this command didn't previously depend on system environment variables right (since those would get wiped away here)?

calufornia (approved):

pkg/deploy/discover/build.go
ameyakhare: changed it to append instead of replace

PR #564, opened 2023-02-27
Add SanitizeParamValues

francescaleung:
## Description

Supports DTS-332.

## Test plan

Unit tests.


pkg/parameters/sanitize.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+			sanitized[k] = sanitizedV
+		} else {
+			sanitized[k] = v
+		}
+	}
+	return sanitized, nil
+}
+
+var secretWords = []string{
```
calufornia, line 48: What about `key` (or at least `API key`)?

calufornia (approved):

PR #567, opened 2023-02-28
Remove Definition interface

colinking:
## Description

Removes the definition interface. We used this when migrating to `Definition_0_3` from earlier definition formats. We have deprecated them and don't have plans to introduce additional changes to this definition format that would require the use of this interface. Removing this interface makes definitions easier to work with by removing one layer of indirection.

## Test plan

Tests should pass.


colinking:
* main
  * **PR #567** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/567" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/567?utm_source=stack-comment).

pkg/deploy/taskdir/definitions/task_def.go
```diff
@@ -895,73 +893,73 @@ func (o *OptionDefinition_0_3) UnmarshalJSON(b []byte) error {
...
-type ScheduleDefinition_0_3 struct {
+type ScheduleDefinition struct {
 	Name        string                 `json:"name,omitempty"`
 	Description string                 `json:"description,omitempty"`
 	CronExpr    string                 `json:"cron"`
 	ParamValues map[string]interface{} `json:"paramValues,omitempty"`
 }
 
-//go:embed schema_0_3.json
+//go:embed task_schema.json
```
francescaleung, line 913: moving this file means we need to change the path in schemastore as well

francescaleung (approved):

pkg/deploy/taskdir/definitions/task_def.go
colinking: Ah, great catch. I reverted this change and added a note about this as a comment. Thank you!

PR #571, opened 2023-03-01
Extract task kinds into separate files

colinking:
## Description

This PR does not include any functional changes. It simply extracts the logic for each task kind into its own `task_def_[...].go` file to slim down `task_def.go`.

## Test plan

Tests pass.

PR #568, opened 2023-02-28
Remove API calls from `definitions` package

colinking:
## Description

This PR removes all API client usage from the `definitions` package. This was only used for fetching resources, which we now pass in directly. This lets the caller decide if an API call is necessary.

These changes are useful in studio, where we perform operations on definitions from within the CLI's API.

## Test plan

Tests should pass.

PR #572, opened 2023-03-01
Add `definitions.Update`

colinking:
## Description

Adds an `Update` method to the `definitions.Definition` struct. This will be used by the CLI (via https://github.com/airplanedev/cli/pull/989) when updating a task via the new config panel.

To keep this simple, I merged it's behavior with the existing `NewDefinitionFromTask` method.

## Test plan

Confirmed I could initialize a task from an existing task, which is the main usage of `NewDefinitionFromTask`.


colinking:
* main
  * **PR #567** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/567" target="_blank">Graphite</a> 
    * **PR #568** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/568" target="_blank">Graphite</a> 
      * **PR #571** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/571" target="_blank">Graphite</a> 
        * **PR #572** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/572" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/572?utm_source=stack-comment).

pkg/deploy/taskdir/definitions/task_def.go
```diff
@@ -46,7 +45,7 @@ type Definition struct {
 
 type taskKind interface {
 	copyToTask(*api.Task, build.BuildConfig, GetTaskOpts) error
-	hydrateFromTask(api.Task, []api.ResourceMetadata) error
+	copyFromTask(api.UpdateTaskRequest, []api.ResourceMetadata) error
```
francescaleung, line 48: why not `api.Task` instead?

francescaleung (commented):
looks fine, but double check your unit tests--if you need to change expected values that seems like a bit of a flag for me

pkg/deploy/taskdir/definitions/task_def.go
colinking: I wanted to make the distinction between `copyToTask` and `copyFromTask` clear...but I agree that's confusing. I'm going to rename it back to `update()` which is what I had it named originally.

colinking:
@francescaleung is there a particular test that you are concerned with? All the test changes are just to initialize previously `nil` values with corresponding empty values. This normalization makes it easier to work with definitions and should be a no-op. I made similar test changes in many of my other related PRs.

pkg/deploy/taskdir/definitions/task_def.go
francescaleung: so you renamed it, but you kept it as `api.UpdateTaskRequest`. original question still stands: why not `api.Task` instead?

francescaleung:
> @francescaleung is there a particular test that you are concerned with? All the test changes are just to initialize previously `nil` values with corresponding empty values. This normalization makes it easier to work with definitions and should be a no-op. I made similar test changes in many of my other related PRs.

Are there places where we rely on stuff being nil or not nil but empty? Does this change JSON/YAML serialization at all? This is mostly my concern.

pkg/deploy/taskdir/definitions/task_def.go
colinking: Because we're updating the definition, so we use an `UpdateTaskRequest` just like we do in the API. It's the same idea as calling `/i/tasks/update` in the API where we apply patch semantic.

colinking:
> Are there places where we rely on stuff being nil or not nil but empty?

Not that I have found. Unfortunately, this is hard to guarantee beyond what we test.

> Does this change JSON/YAML serialization at all?

No, because an empty map/array is considered "empty" just like `nil` when serializing a definition. We don't rely on serializing `null` values for definitions (we use `omitempty`).

francescaleung (approved):

PR #574, opened 2023-03-02
Add tracing functionality to lib

yolken-airplane:
## Description
This change adds some basic tracing functionality to lib so that we can use it across all of our golang repos including airport and cli. It's adapted from the code in airport, but generalized a bit so that it doesn't require the airport authn libraries.

The tracer provider functionality is not included since that will be application-specific. The idea is that cli, airport, etc. will handle tracing initialization themselves.

Once this is in, I'll have a few follow-up changes to use this library in both airport and cli.

## Test plan
Testing completed successfully via updating the lib version of the cli and adding some basic tracing functionality in `airplane dev`. 

Here's an example trace: https://ui.honeycomb.io/airplane/datasets/stage/result/cZa7ppqxEo5/trace/wssRNFyHTWD?fields[]=c_name&fields[]=c_service.name&span=d605fab95e33fa19

Screenshot 2023-03-01 at 7 13 35 PM

I also tested using baggage for the authn fields in airport, and that worked as expected.


.editorconfig
```diff
@@ -1,9 +1,9 @@
 root = true
 
-[*]
+[*.{js,py,ts}]
```
yolken-airplane, line 3: @colinking has a fix for the golang formatting here, I'll revert this once that's in.

pkg/utils/tracing/tracing.go
```diff
@@ -0,0 +1,202 @@
...
+		context.Background(),
+		resource.WithProcess(),
+		resource.WithOS(),
+		resource.WithHost(),
+		resource.WithAttributes(
+			semconv.ServiceNameKey.String("cli"),
+			attribute.String("authn_team_id", opts.TeamID),
+			attribute.String("authn_user_id", opts.UserID),
+			attribute.String("authn_user_email", opts.UserEmail),
+			attribute.String("cli_version", opts.Version),
```
leeweisberger, line 191: Can airport take advantage of this (it also imports lib)? Does it just leave this one blank? What about the service name?

leeweisberger (changes requested):
Can airport use this as well?

yolken-airplane:
> Can airport use this as well?

Possibly. There are few significant differences, including the tracer initialization and also the `getDefaultSpanFields` field- in the airport case, it calls some airport-specific functionality to set the user and team. 

Let me think about this some more. If we can get around the `getDefaultSpanFields` issue, then maybe we can take the tracer initialization out of here and just have each app do its own initialization.

yolken-airplane:
Actually, I think we can use the baggage to set the authn stuff in airport. Let me see if that works...

yolken-airplane:
> > Can airport use this as well?
> 
> Possibly. There are few significant differences, including the tracer initialization and also the `getDefaultSpanFields` field- in the airport case, it calls some airport-specific functionality to set the user and team.
> 
> Let me think about this some more. If we can get around the `getDefaultSpanFields` issue, then maybe we can take the tracer initialization out of here and just have each app do its own initialization.

Ok, I've done some cleanups so that this library can be used in airport too. We can use the baggage-related functionality to get around not having access to the airport authn libraries in `getDefaultSpanFields`.

@leeweisberger can you take another look when you get a chance?

leeweisberger (approved):

PR #579, opened 2023-03-02
Add event listeners for errors and unhandledrejections to send messages to parent frame

rose-huang:
## Description
We add event listeners when we get an error event or unhandledrejection event in the view

We duplicate some code here that we have in the views lib (ViewMessage and sendViewMessage) but this seems better than trying to export those out of views and having to deal with backward compatibility issues

We also start sending the stack trace of the error so we can show that in the debug panel for errors that don't have a modal

## Test plan
Testing alongside views and web changes


pkg/build/views/main.tsx
```diff
@@ -82,13 +84,66 @@ for (const [key, value] of getAllQueryParams()) {
...
+// Intercept window error and unhandledrejection events and send them to the parent window.
+if (typeof window !== undefined) {
+  window.addEventListener("error", (event) => {
+    sendViewMessage({
+      type: "console",
+      messageType: "error",
+      message: "Uncaught Error: "+event.error.message,
+      stack: event.error.stack,
+      hash: hash(event.error),
+      time: dayjs().unix(),
```
leeweisberger, line 103: Same here. No dayjs

pkg/build/views/package.json
```diff
@@ -9,11 +9,14 @@
   "dependencies": {
     "@airplane/views": "1.14.4",
     "react": "18.2.0",
-    "react-dom": "18.2.0"
+    "react-dom": "18.2.0",
+    "object-hash": "3.0.0",
```
leeweisberger, line 13: This doesn't actually provide this dependency, it's just a place where we track dependency types. We need to update the view build to ensure that object-hash is installed.

Look up GenViewsShimPackageJSON. You need to add the object-hash there.

 Also update renovate.json so that we don't update this dependency often

leeweisberger (changes requested):

pkg/build/views/package.json
rose-huang: I still keep this here right? in addition to adding to `GenViewsShimPackageJSON`?

pkg/build/views/package.json
```diff
@@ -9,11 +9,13 @@
   "dependencies": {
     "@airplane/views": "1.14.4",
     "react": "18.2.0",
-    "react-dom": "18.2.0"
+    "react-dom": "18.2.0",
+    "object-hash": "3.0.0"
```
leeweisberger, line 13: Yup we still keep this here. It helps us keep these deps up to date because renovate automatically looks at package.json files and updates them, That's the only reason we're using a package.json

leeweisberger (approved):

PR #599, opened 2023-03-16
test

calufornia:


francescaleung:
can you add a test for this :)

francescaleung:
also I think this produces a REST resource & not a graphql resource, that's no bueno

calufornia:
> also I think this produces a REST resource & not a graphql resource, that's no bueno

Oh good point, let me fix that

pkg/resources/kinds/graphql_test.go
calufornia, line 1: I basically just copied this from `rest_test.go`

pkg/resources/factory_test/factory_test.go
```diff
@@ -45,6 +46,35 @@ func TestResourceFactories(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "graphql",
```
calufornia, line 50: This test actually catches the bug

calufornia:
@francescaleung Updated, feel free to approve if it looks good or just take it over 😂

francescaleung (approved):

PR #604, opened 2023-03-17
Add object-hash to build deps

calufornia:
## Description

This is also a dep for developing locally.

## Test plan
n/a


leeweisberger (approved):
Why are views working without this...??

leeweisberger:
> Why are views working without this...??

Shouldn't this crash views that hit this part of the code? Maybe it's because it's pretty rare to hit this?

calufornia:
> > Why are views working without this...??
> 
> Shouldn't this crash views that hit this part of the code? Maybe it's because it's pretty rare to hit this?

I'm...not entirely sure. I'm surprised installing the dependencies even works given `"object-hash"` was previously going to be set to any empty string: https://github.com/airplanedev/lib/blob/d0341d42ba1a27e6187c6d88d2d73f90720f1302/pkg/build/view.go#L74

cc @rose-huang

PR #590, opened 2023-03-14
Infer JS transformer serialization from definition

colinking:
## Description

This PR refactors the JS transformer to infer the JS SDK's serialization format from the JSON-based definition format. This allows us to add new definition fields without having to tweak the JS transformer (in most cases). We instead do some light transformation to convert the JSON format into the JS SDK format (e.g. converting parameter values into their JavaScript-y equivalents, such as `Date` objects).

This change also adds support for restrict callers to the JS transformer.

## Test plan

Unit tests should pass.

PR #592, opened 2023-03-14
Support editing YAML-based task definitions

colinking:
## Description

This PR adds support for updating YAML definitions. The logic is simplistic, simply relying on the existing YAML serialization. In the future, we could replace this implementation with an AST-based implementation that can do fancier logic such as retaining comments.

Resolves DTS-340

## Test plan

Unit tests pass.

PR #591, opened 2023-03-14
Add additional tests to JS transformer

colinking:
## Description

Extends the test coverage on the JS transformer. Adds support for serializing parameters as strings (instead of objects) if only the type is set. Fixes quote formatting s.t. single/double quotes are chosen to minimize string length.

## Test plan

Unit tests pass.

PR #606, opened 2023-03-20
Various task definition improvements

colinking:
## Description

Unwraps the `timeout` and `resources` fields into `int` and `map` values, respectively. This makes it easier to construct a `Definition` struct. The structs are no longer necessary (e.g. there used to be some default timeout behavior but that was removed).

Marks a few fields in the JSON schema as non-required. These haven't been required by our API, but for some reason were marked as required in the schema.

Tweaks the serialization of options to not use label objects when not needed. This is useful in the JS transformer (see upstream).

For more context, see comments in the PR these changes were extracted from: https://github.com/airplanedev/lib/pull/590

## Test plan

I tested:
1. I can deploy a task with resources, options, and timeouts
2. I can initialize an inline and YAML-based JS task
3. The studio can discover/dev inline JS tasks


colinking:
* main
  * **PR #606** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/606" target="_blank">Graphite</a>  👈
    * **PR #590** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/590" target="_blank">Graphite</a> 
      * **PR #591** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/591" target="_blank">Graphite</a> 
        * **PR #592** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/592" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/606?utm_source=stack-comment).

pkg/runtime/javascript/fixtures/transformer/parameters.out.airplane.js
```diff
@@ -11,7 +11,7 @@ export default airplane.task(
         required: false,
         default: "My default",
         regex: "^.*$",
-        options: ["Thing 1", "Thing 2", {
```
ameyakhare, line 14: why is this removed? we should be able to have options without labels right?

ameyakhare (approved):

pkg/runtime/javascript/fixtures/transformer/parameters.out.airplane.js
colinking: Because of the change to option marshaling, it broke that specific JS transformer test. I didn't bother to backport the fix from the upstream PR since it would have created a bunch of merge conflicts and the JS transformer isn't used by customers yet.

PR #610, opened 2023-03-22
Default to top level configs for attachments

ameyakhare:
## Description
After we deprecated nested configs in rest/sql tasks (defaulting to the top level config), we stopped defaulting to the top level configs for config attachments. This fixes that. This makes sure that sql/rest tasks that are defined with top level configs and use configs in JSTs actually work.

## Test plan
When I use this change for running a local studio task that has configs attached at the top level OR nested it works properly now.


francescaleung (approved):
can you write a test for this?

PR #615, opened 2023-03-23
Fix workflows ARM builds

yolken-airplane:
## Description
This change updates the lib node builder to not delete the temporalio-core aarch64 binaries. These are needed for ARM workflow builds and runs.

The images will be slightly larger without this deletion, but it's probably better to avoid platform-related problems here.

Resolves DTS-491

## Test plan
Testing completed successfully via updating the builder and running a workflow build on Mac and verifying that everything worked as expected.


pkg/build/node.go
```diff
@@ -1283,7 +1283,7 @@ func makeInstallCommand(req makeInstallCommandReq) string {
 		installCommand = "npm ci"
 	}
 	// Remove large binaries for platforms that we aren't using
-	installCommand += " && rm -Rf /airplane/node_modules/@swc/core-linux-x64-musl /airplane/node_modules/@temporalio/core-bridge/releases/aarch64* /airplane/node_modules/@temporalio/core-bridge/releases/*windows* /airplane/node_modules/@temporalio/core-bridge/releases/*darwin*"
+	installCommand += " && rm -Rf /airplane/node_modules/@swc/core-linux-x64-musl /airplane/node_modules/@temporalio/core-bridge/releases/*windows* /airplane/node_modules/@temporalio/core-bridge/releases/*darwin*"
```
leeweisberger, line 1286: Is it worth it to do this conditionally, only if we aren't doing an arm build? How large are these?

leeweisberger (approved):
Should we do this conditionally?

pkg/build/node.go
yolken-airplane: They're 15-20MB each, so shouldn't hurt image size too much?

leeweisberger: ok let's just not worry about it.

yolken-airplane: Let me see how easy it would be to base it on the platform...

yolken-airplane: > ok let's just not worry about it.

Yeah, SGTM. I started playing around with it but it's kind of a pain because the output of `uname -m` can vary slightly from environment to environment (in docker vs. not in docker, etc.). Added a TODO for now.

PR #618, opened 2023-03-28
Support updating SQL queries

colinking:
## Description

Adds support for editing SQL queries when updating tasks.

Resolves DTS-495

## Test plan

Unit tests pass.


colinking:
* main
  * **PR #618** <a href="https://app.graphite.dev/github/pr/airplanedev/lib/618" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/lib/618?utm_source=stack-comment).

pkg/runtime/sql/sql.go
```diff
@@ -86,15 +79,54 @@ func (r Runtime) FormatComment(s string) string {
...
-	return updaters.CanUpdateYAML(path)
+	if canUpdate, err := updaters.CanUpdateYAML(path); err != nil {
+		return false, err
+	} else if !canUpdate {
+		return false, nil
+	}
+
+	if _, err := os.Stat(path); err != nil {
+		return false, errors.Wrap(err, "opening file")
+	}
```
francescaleung, line 129: does this error if the file doesn't exist? is that desirable behaviour?

pkg/runtime/sql/sql.go
```diff
@@ -86,15 +79,54 @@ func (r Runtime) FormatComment(s string) string {
...
 	return true
 }
 
 func (r Runtime) Update(ctx context.Context, logger logger.Logger, path string, slug string, def definitions.Definition) error {
-	return updaters.UpdateYAML(ctx, logger, path, slug, def)
+	if err := updaters.UpdateYAML(ctx, logger, path, slug, def); err != nil {
+		return err
+	}
+
+	// Update the SQL query within the entrypoint file.
```
francescaleung, line 91: if this fails the yaml is still updated, right? How hard is it to make it more atomic?

francescaleung (approved):

pkg/runtime/sql/sql.go
colinking: Yeah, an error seems reasonable. It might be worth handling this case with a 400 specifically (e.g. "your definition file is invalid"), but I'm going to leave as-is for now.

pkg/runtime/sql/sql.go
colinking: We can get marginally better by checking that the file exists first before updating the YAML file, but since we don't handle that with a 400 (as above), I'll punt on that until then.

